PR #2, opened 2015-07-10
Derive eq

ceedubs:
This is copy/pasted from https://github.com/ceedubs/shapeless-cats.


non:
:+1:

milessabin:
What `Eq` instances do we have without derivation?

ceedubs:
@milessabin `CatsSuite` is extending [`std.AllInstances`](https://github.com/non/cats/blob/master/std/src/main/scala/cats/std/all.scala#L4), so we are picking up a bunch for `Int`, `String`, `List`, etc.

milessabin:
Cool ... could you check that in the derivation of a complex instance that the handwritten ones are used (eg. an explicit instance for `List` or `Option` should be used for an embedded `List` or `Option` instead of being derived afresh.

Also, would you mind changing the import convention to match `Functor`, `Foldable`, `Iterable` etc. ie. `import cats.derived.functor._` etc. ...

ceedubs:
@milessabin I originally meant to add a test that the handwritten instances are used, but I must have gotten lazy. I will update this PR with those tests in the near future.

Sure, I'll change the import convention as well.

ceedubs:
Sorry for the delay. I haven't forgotten about this, but I've gotten quite busy. I hope to get to it this week, because if I don't I probably won't be able to get to it for about a month.

milessabin:
I'm mainly focussed on the higher minded type classes right now, so I'm happy to leave this on the offchance that you'll be able to snatch a moment or two for it.

ceedubs:
> higher minded

Miles, I know you have done some impressive things with Scala, but there are limits; if only for ethical reasons.

milessabin:
Hmm ... I do seem to be getting a little above myself :-D

ceedubs:
@milessabin I changed `EqDerivedOrphans` to `eq` to match the existing import convention.

I added a test that existing instances in the implicit scope are respected, but the test fails :\. My local (and terrible) `Eq[Int]` instance doesn't seem to be used. I would expect this to either work or to result in a compile error about ambiguous `Eq[Int]` instances, since one is also inherited from `std.AllInstances`. Am I doing something wrong here?

ceedubs:
As I mentioned the other day, I'll be mostly out of commission for the next month. @milessabin please feel free to make any changes you see fit if you want to merge this.

ceedubs:
@milessabin do you have any thoughts on https://github.com/milessabin/kittens/pull/2#issuecomment-124896757 ?

milessabin:
@ceedubs sorry it's taken so ridiculously long for me to get back to this :-(

I think the failure you're seeing there is expected under pretty much all reasonable interpretations. The `arbFoo` instance is accessible without prefix in the test, so it should be selected in preference to anything derived. TBH, I think this is quite a good example of why we really shouldn't be importing instances and should instead be injecting them into the implicit scope using export-hook, although even then I would expect the normal priority to prefer an explicit instance for `Foo` over a derived one.

I'm going to fiddle around with a little and get it into the current canonical form. I'll update the test with something which validates the normal prioritization and also something which shows that your derived instance can be made the preferred one.

milessabin:
Merged with the tweaks here: 19ecbc218c70c2433bd601c24621539bd7996140.

PR #3, opened 2016-01-12
ported sequence from shapeless.contrib.scalaz here.

kailuowang:
May this be the first step of porting scalaz-shapeless utils here.

### DIFF
--- a/core/src/main/scala/cats/sequence/package.scala
+++ b/core/src/main/scala/cats/sequence/package.scala
@@ -1,0 +1,2 @@
+package cats
+package object sequence extends SequenceOps
--- a/core/src/main/scala/cats/sequence/sequence.scala
+++ b/core/src/main/scala/cats/sequence/sequence.scala
@@ -1,0 +1,79 @@
+/*
+ * Adapted from shapeless-contrib scalaz
+ * https://github.com/typelevel/shapeless-contrib/blob/v0.4/scalaz/main/scala/sequence.scala
+ *
+ */
+package cats.sequence
+
+import shapeless._
+
+import cats._
+
+import scala.annotation.implicitNotFound
+
+trait Apply2[FH, OutT] {
+  type Out
+  def apply(fh: FH, outT: OutT): Out
+}
+
+object Apply2 {
+  type Aux[FH, OutT, Out0] = Apply2[FH, OutT] { type Out = Out0 }
+
+  implicit def apply2[F[_], H, T <: HList](implicit app: Apply[F]): Aux[F[H], F[T], F[H :: T]] =
+    new Apply2[F[H], F[T]] {
+      type Out = F[H :: T]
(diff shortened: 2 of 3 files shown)


kailuowang:
when testing on, scala 2.10, travis complained no output in 10 minutes and terminated the build. Can you try re-run it?

ceedubs:
This looks good to me.

`ValidationNel` is common enough in this sort of use-case that it might be nice to have a test for it.

kailuowang:
@ceedubs added test for `ValidationNel` although as in scalaz, ValidationNel type alias breaks implicit search.
I also had to write 
```
implicit def nelSemiGroup[A]: Semigroup[NonEmptyList[A]] = implicitly[SemigroupK[NonEmptyList]].algebra[A]
```
I am curious why it's needed, that is, why the following function doesn't work
```
implicit def nelSemiGroup[A, F[_]](implicit sgk: SemigroupK[F]): Semigroup[F[A]] = sgk.algebra[A]
```

milessabin:
This looks good to me apart from the package name. Can we come up with something a little less tied to a implementation detail? Preferably not `cats.util` ;-)

Also, have you considered a variant of sequence defined in terms of `ProductArgs`? That would allow applications of the form `sequence(x, y, z)` as an alternative to `sequence(x :: y :: z :: HNil)`.

kailuowang:
@milessabin I see this package playing the same role as the shapeless.contrib.scalaz. how about `cats.shapeless` ? 
`ProductArgs` sounds a great idea. I also need to add a version for record, and I can use `RecordArgs` there. I have other coworkers interested in contributing to this package. So the question is that do you want to have this PR merged (with the new package name) so that other people can start adding stuff, or do you want to wait until I finish the `record` and `ProductArgs` features?

core/src/main/scala/cats/hlist/sequence.scala
```diff
@@ -0,0 +1,90 @@
+/*
+ * Adapted from shapeless-contrib scalaz
+ * https://github.com/typelevel/shapeless-contrib/blob/master/scalaz/main/scala/sequence.scala
```
dwijnand: Don't link to master as that URL that file might not always be at that path.

How about `https://github.com/typelevel/shapeless-contrib/blob/v0.4/scalaz/main/scala/sequence.scala`

dwijnand:
`cats.auto`? `cats.seq`?

kailuowang:
@milessabin renamed package to cats.generic and added the ProductArgs and an implicit pimp for HList.

milessabin:
I think `cats.generic` is a little too general. Can we go with `cats.sequence`?

kailuowang:
renamed to cats.sequence

milessabin:
LGTM! Many thanks ... merging :-)

PR #4, opened 2016-01-15
Added sequence for record and generic case classes 

kailuowang:
example usage
```
  import cats.sequence._
  import cats.implicits._

  case class MyCase(a: Int, b: String, c: Float)

  val myGen = sequenceGeneric[MyCase]

  myGen(a = Option(1), b = Option("23"), c = Option(2.2))  //gives Some(MyCase(1, "23", 2.2))
  
```

Please help review the code. I had to duplicate between Functor[F[_]] and Functor[F[A, ?]], any advice on how to reduce such duplication?

### DIFF
--- a/core/src/main/scala/cats/sequence/sequence.scala
+++ b/core/src/main/scala/cats/sequence/sequence.scala
@@ -8,6 +8,8 @@ package cats.sequence
 import shapeless._
 
 import cats._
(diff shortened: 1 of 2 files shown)


milessabin:
I'm rerunning the Travis build for 2.10.x now ...

milessabin:
LGTM. Many thanks ... merging!

PR #7, opened 2016-01-18
refactor to eliminate a different sequence ops name for record

kailuowang:
based on @milessabin 's SO answer to this [question](http://stackoverflow.com/questions/34846947/how-to-create-overloads-for-product-and-record)


milessabin:
Awesome! Thanks!

I'll merge this as is, but would you mind sending another PR with the white space matching the conventions in the rest of the project (roughly: not vertical alignment, two space indents, break lines aggressively).

kailuowang:
Will do thanks

PR #9, opened 2016-01-19
added support for Kleisli and Monad Transformer with 3 type parameters, in sequence

kailuowang:


milessabin:
More good stuff! Many thanks ... merging :-)

PR #13, opened 2016-03-04
make util typeclasses serializable

kailuowang:


milessabin:
That looks like the same record type problem we had with 2.10.x previously.

kailuowang:
fixed

milessabin:

PR #14, opened 2016-03-07
For review only (do not merge yet) - added an applicative based on pure foldable and functor

kailuowang:
trampolining doesn't seem to work, @milessabin can you take a look?
another thing I am not sure is how to do reexport, since this Applicative requires derived Functor while itself also implements Functor, import both derived.applicative._ and dervied.functor._ causing ambiguous implicits problem, thus I don't know how to achieve auto derivation of `Applicative`.

### DIFF
--- a/core/src/main/scala/cats/derived/applicative.scala
+++ b/core/src/main/scala/cats/derived/applicative.scala
@@ -1,0 +1,41 @@
+package cats.derived
+
+import cats.{Foldable, Functor, Eval, Applicative}, Eval.now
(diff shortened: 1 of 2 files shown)


kailuowang:
Also this approach seems to be taking quite some time to compile, which is probably why Travis timed out. I don't quite understand then reason for slow compilation either since this derivation is mostly just a delegation to three other derivations.

kailuowang:
closing this one, will submit another one later.

PR #15, opened 2016-03-08
Added the ability to derive Monad and Apply

kailuowang:
Notes to @milessabin Trampoline works properly (It was an error in the test.)
Exporting doesn't work yet. Partly due the fact that since this Apply requires derived Functor while itself also implements Functor, import both derived.apply._ and dervied.functor._ causing ambiguous implicits problem, thus I don't know how to achieve auto derivation. 
For monad, compiler hangs when `importing both monad and foldable`..

### DIFF
--- a/core/src/main/scala/cats/derived/apply.scala
+++ b/core/src/main/scala/cats/derived/apply.scala
@@ -1,0 +1,58 @@
+package cats.derived
+
+import cats.{Foldable, Functor, Eval, Apply}, Eval.now
+import alleycats.{EmptyK, Pure}
+import export.{ exports, imports, reexports }
+import shapeless._
+
+// todo doesn't work because an implicit Functor needs to be scope while Apply itself implements Functor
+//@reexports[MkApply]
+//object apply {
+//  @imports[Apply]
+//  object legacy
+//}
+
+trait MkApply[F[_]] extends Apply[F]
+
+@exports
+object MkApply extends MkApply0 {
+  def apply[F[_]](implicit maf: MkApply[F]): MkApply[F] = maf
+}
+
+trait MkApply0 extends MkApply1 {
+  implicit def mkWithEmptyK[F[_]](
+     implicit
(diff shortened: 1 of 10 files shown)


kailuowang:
This two new type classes significantly increase the compilation time. (locally it took 5 minutes to do an incremental compilation when one test file is changed). Although they themselves don't do any derivation.

milessabin:
Do you have any insight into what might be going wrong with 2.10.x?

kailuowang:
Not really. Will give it a look today.

kailuowang:
Pushed a fix that isolates the derivation of Foldable and Functor in objects. Not sure if that's the best way to address the problem.

kailuowang:
I think I found a solution for 2.10.x on travis. But not ready for merging yet, I'll make some other adjustment and squash.

core/src/main/scala/cats/derived/apply.scala
```diff
@@ -0,0 +1,58 @@
...
+     Fdb: Cached[Foldable[F]]
+  ): MkApply[F] = {
+    val nullEmpty = new EmptyK[F] {
+      def empty[A]: F[A] = null.asInstanceOf[F[A]]
+    }
+    doMkWithEmptyK(nullEmpty, F.value)
+  }
+
+  protected def doMkWithEmptyK[F[_]](E: EmptyK[F], F: Functor[F])
+                       (implicit
```
milessabin: Whitespace? ;-)

milessabin:
Great work! Thank you :-)

kailuowang:
@milessabin ready for review and merge. 
Two notes: 
1) tests for monad and applicative now lives in its own module, they just take too long to compile will significantly slow down the development process. 
2) using `Cached` in multiple derivation methods definitely helps compilation time (basically this way you don't need to re-attempt to get a typeclass instance if it's already cached by another derivation method.)

milessabin:
This is great stuff! Many thanks :-)

PR #16, opened 2016-03-19
added a semigroup for product

kailuowang:
simple implementation for product, but IMO should be pretty useful.


milessabin:

PR #19, opened 2016-05-01
no more work arounds for si-2712

kailuowang:


kailuowang:
just realized that this would be a breaking change for use site. close for now.

PR #20, opened 2016-05-02
SI2712 fix + applicative

kailuowang:
update: 
Additional notes. This is a simple fix that
1,  moves si-2712 work arounds for `sequence` implicits to priority implicit traits, and
2, provides an `Applicative` derivation simply from `Pure` and `Apply` since both can be also derived.


milessabin:

PR #22, opened 2016-09-08
Introduce generalization for "Copair" types like Either, Xor, Validated, A :+: B :+: CNil

Jacoby6000:
There has been some discussion on this already over in the cats repo https://github.com/typelevel/cats/pull/976.  

Some people suggested that I submit this over here instead.


milessabin:
This is great stuff, but I don't think it belongs in Kittens. It looks very much along the lines of what I have planned for shapeless 3.0.

PR #23, opened 2016-09-08
update to cats 0.7.2

kailuowang:
looks like we are having an issue with `cats.instances.tuple._`, so I needed to replace `import cats.implicits._` with more specific imports.


non:
Assuming things are green 👍 from me.

kailuowang:
Passed on 2.10 but failed on 2.11, Travis hates kittens.

milessabin:
LGTM :-)

PR #24, opened 2016-09-17
removed unecessary sequencer mk

kailuowang:
fixed the issue with cats tuple instance https://github.com/typelevel/cats/issues/1368


milessabin:
LGTM ... many thanks! Merging :-)

PR #25, opened 2016-10-20
Added Monoid derivation for products

joroKr21:


kailuowang:
Thanks very much @joroKr21. 
It failed on scala 2.10 build, but I suspect if you give it a bit nudge by providing the Eq instance in the companion object explicitly. That being said, 2.10 has been giving us quite some headaches. We are trying to get some consensus on if we still need to support 2.10 when 2.12 is around the corner. 
Another issue is this implementation has some duplication with the `Semigroup` one, another approach might be deriving instance for `Empty` and then combine the two into `Monoid`. However, that approach might have lower compilation performance, as it turned to be the case in `Monad` derivation code. So I am okay with the approach you took .

joroKr21:
I see, I'll add an `Eq[Foo]` instance explicitly. Btw, it may be a good idea to move `Foo` outside, since it's used in other tests as well.

I was wondering about `Empty`, but I looked for it in the wrong place (`cats` instead of `alleycats`). However, now that I tried to do it this way (joroKr21@0b0ac12), the test blows up with a StackOverflow. Is it because `Monoid` extends `Semigroup` and `Empty` has `monoidIsEmpty` thus causing circular implicits?

Another thing, should I add `@exports(Subclass)` to `MkMonoid` to avoid conflicts with `MkSemigroup`?

kailuowang:
That might be a circular derivation, though I am not any more certain than you. 
As for avoiding `MkMonoid` conflicting with `MkSemigroup`, I think kittens users should probably only use a la carte import, so I am not sure if it's necessary.

kailuowang:
👍

milessabin:
Great stuff! Thanks :-)

Would you mind adding your details to CONTRIBUTORS.md before I merge?

joroKr21:
I realized we can workaround the circular implicits problem by using the `MkEmpty` and `MkSemigroup` typeclasses instead of `Empty` and `Semigroup` which is of course more restrictive, but equivalent to the standalone implementation (see joroKr21@fcd164b).

So the question is what is more important - having less code duplication (algebraic version) or generating fewer instances (standalone version). I guess the same thing applies for deriving `MonoidK`.

milessabin:
Let's start with `MkEmpty` and `MkSemigroup` and see how it goes from there.

milessabin:
Is this ready for review now?

joroKr21:
Yes, it should be ready.

PR #26, opened 2016-11-04
Added SemigroupK and MonoidK derivation

joroKr21:
I took the standalone approach in `MkMonoidK` for two reasons:

1. Faster compilation
2. `MkEmptyK[O[I[?]]]` and `MkSemigroupK[O[I[?]]]` can derive instances that taken together don't form a `MonoidK[O[I[?]]]`. E.g. consider `NonEmptyList[Option[?]]`

### DIFF
--- a/core/src/main/scala/cats/derived/monoidk.scala
+++ b/core/src/main/scala/cats/derived/monoidk.scala
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2015 Miles Sabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


milessabin:
Is this ready for review now?

joroKr21:
Yes, I think it's ready for review.

core/src/main/scala/cats/derived/monoidk.scala
```diff
@@ -0,0 +1,89 @@
...
+      def combineK[A](x: F[A], y: F[A]) =
+        pack(fo.map2(unpack(x), unpack(y))(fi.combineK(_, _)))
+    }
+}
+
+trait MkMonoidK2 extends MkMonoidK3 {
+  implicit def generic[F[_]](implicit gen: Generic1[F, MkMonoidK])
+    : MkMonoidK[F] = new MkMonoidK[F] {
+      import gen._
+      def empty[A] = gen.from(fr.empty)
```
kailuowang: nit here: since you already imported `gen._` , `from` will be enough as below right?

milessabin:
Do you want to fix the nit or shall I merge as is?

joroKr21:
Fixed it :smile:

milessabin:
Many thanks! Merging :-)

PR #27, opened 2016-12-13
updated scala.js and sbt release

kailuowang:


kailuowang:
not sure I understand why travis broken with 

> :::: WARNINGS
> 		module not found: org.scala-lang#scala-library;2.10.6
> 

I was able to download that jar from the url mentioned in the error logs.

kailuowang:
alright, retried the 3rd time and it's a charm.

PR #35, opened 2017-03-04
SBT 0.13.13-M1 -> 0.13.13, CrossVersion.full -> CrossVersion.patch for Typelevel Scala compatibility.

ClydeMachine:
Addressing the issue described at [typelevel/scala#135](https://github.com/typelevel/scala/issues/135). Moved to SBT version 0.13.13 from 0.13.13-M1, set CrossVersion.patch from CrossVersion.full, and set scalaOrganization.value from "org.scala-lang" in library dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,10 +114,10 @@ using Typelevel Scala and see if it resolves the issue.
 
 To use Typelevel Scala you should,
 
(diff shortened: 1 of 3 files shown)


milessabin:
LGTM! Many thanks ... merging :-)

PR #36, opened 2017-04-13
 removed unapply machinery

kailuowang:
Standardize towards partial unification fix for SI-2712. I was able to reduce the lines of code by 40%.


milessabin:
Awesome stuff ... thanks!

ceedubs:
Very nice!

PR #37, opened 2017-05-02
copied trivial from cats.core

kailuowang:


core/src/main/scala/cats/derived/trivial.scala
```diff
@@ -20,3 +20,27 @@ trait Trivial1[F[_]]
...
+ * companion object, covering a few common use cases and avoiding the
+ * need for unnecessary lambdas (e.g. if you want a trivial type class
+ * instance for a type constructor, you should use `Trivial.PH1`).
+ *
+ * Moved from cats.core
+ */
+sealed trait Trivial
+
+object Trivial {
+  type P1[A] = Trivial
```
milessabin: What's the motivation for these aliases?

core/src/main/scala/cats/derived/trivial.scala
```diff
@@ -20,3 +20,27 @@ trait Trivial1[F[_]]
...
+  type PH1[F[_]] = Trivial
+  type P1H1[F[_], A] = Trivial
+  type P2[A, B] = Trivial
+  type P2H1[F[_], A, B] = Trivial
+  type P3[A, B, C] = Trivial
+  type P3H1[F[_], A, B, C] = Trivial
+
+  implicit val catsTrivialInstance: Trivial = new Trivial {}
```
milessabin: I think `mkTrivial` would be better here, for consistency with `Trivial1`.

milessabin:
OK, looking more closely, `Trivial.PH1` replaces `Trivial1` ... can we motivate the naming of the aliases and ditch `Trivial1` everywhere?

kailuowang:
I replaced Trivial1 and renamed the instance creation method. By "motivate the naming of the aliases", did you mean naming them in a way that shows motivation?

milessabin:
Well I'd like to know what the current set of names are meant to convey ... there's a pattern there, but it's not clear what it is or what the choice of letters signifies.

kailuowang:
I had the same feeling, but I struggle to come up with a naming scheme myself. pinging @djspiewak

djspiewak:
I'm guessing I meant "partial higher-order position arity-1", but that's just a guess. If the alias isn't obvious, it's a bad alias.

kailuowang:
Closing due to lack of interest

PR #39, opened 2017-06-16
Add 2.12.2 to .travis.yml

macalinao:


milessabin:
I think you'll need to bump the Scala.js version as well.

macalinao:
good point @milessabin -- fixed! :)

milessabin:
LGTM ... many thanks ... merging :-)

milessabin:
Ahh ... actually, I missed that you hadn't also updated build.sbt ... would you mind doing that as well?

macalinao:
whoops... i'll do that. Should I remove 2.11.1 from .travis.yml as well? I didn't think of this as a version upgrade, just testing to see if tests work under 2.12.2.

milessabin:
Yes please! And would you mind bumping 2.11.x to 2.11.11 everywhere as well? :-)

PR #42, opened 2017-07-13
Don't use the catch-all cats dependency

travisbrown:
We're thinking of using this and pulling in cats-laws etc. at runtime isn't ideal. I've also updated some build and test plugins while I was here.


kailuowang:
LGTM Thanks @travisbrown

milessabin:
Also LGTM ... many thanks ... merging :-)

PR #46, opened 2017-08-04
Post release update

kailuowang:


kailuowang:
close and reopen to restart the build

kailuowang:
I don't know what's wrong with that build, which simply wouldn't work with git. I am going to merge anyway since it's a straight version bump in readme.

PR #47, opened 2017-08-07
WIP Add Show derivation

### LINKED ISSUE #40: cats.derived.show
it would be useful to have typeclass derivation for `cats.Show` for (labelled) products and coproducts. I guess even unlabelled products/coproducts could be useful too but I have no usecase for that.

LukaJCB:
Doesn't quite work yet, might need some help getting the first test to pass. Should resolve #40

### DIFF
--- a/core/src/test/scala/cats/derived/show.scala
+++ b/core/src/test/scala/cats/derived/show.scala
@@ -1,0 +1,31 @@
+package cats.derived
+import cats.Show
+import cats.derived.TestDefns.{Foo, Node, Tree, Leaf}
+import MkShow._
+
+class ShowTests extends KittensSuite {
+
+  test("for simple product") {
+    import cats.instances.all._
+    val foo = Foo(42, Option("Hello"))
+    assert(Show[Foo].show(foo) == "Foo(i = 42, b = Some(Hello))")
+  }
+
+  test("for tree") {
+    import cats.instances.all._
+    val tree = Node(Leaf(1), Node(Node(Leaf(2), Leaf(3)), Leaf(4)))
+    assert(Show[Tree[Int]].show(tree) ==
+      "Node(l = Leaf(t = 1), r = Node(l = Node(l = Leaf(t = 2), r = Leaf(t = 3)), r = Leaf(t = 4)))")
+  }
+
+  test("existing Show instances in scope are respected") {
+
+    implicit def showInt: Show[Foo] = new Show[Foo] {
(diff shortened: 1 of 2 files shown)


core/src/test/scala/cats/derived/show.scala
```diff
@@ -0,0 +1,31 @@
...
+import cats.Show
+import cats.derived.TestDefns.{Foo, Node, Tree, Leaf}
+import MkShow._
+
+class ShowTests extends KittensSuite {
+
+  test("for simple product") {
+    import cats.instances.all._
+    val foo = Foo(42, Option("Hello"))
+    assert(Show[Foo].show(foo) == "Foo(i = 42, b = Some(Hello))")
```
LukaJCB: This currently only prints `"i = 42, b = Some(Hello)"`

Any pointers?

milessabin: It's the coproduct instances which are responsible for rendering which branch of the coproduct you're traversing to, so if you start at the product then, as here you'll just get a rendering for the content of the product. Try adding a `Typeable` instance at the product level and use it's `describe` method.

SystemFw: This can't work for `Tree`

kailuowang: @milessabin @LukaJCB what are you thoughts on this? I think @SystemFw's solution, although not perfect but a great addition to kittens. It would be nice merge and release kittens with some `Show` derivation.

SystemFw: Let me know if I need to PR it or not

kailuowang: @SystemFw, A PR would be really appreciated.

SystemFw: @kailuowang I will open one over the weekend

LukaJCB: Yes thank you @SystemFw, I guess it is not ideal, but it will work for a lot of cases. @milessabin said on gitter, we might be able to provide full derivation by splitting `describe` from `Typeable`. Should I open an issue on this? :)

SystemFw: @LukaJCB https://github.com/milessabin/shapeless/issues/750

SystemFw:
This should probably be closed in light of https://github.com/milessabin/kittens/pull/48

PR #48, opened 2017-08-12
Add generic derivation for Show (solves #40 with a known limitation)

SystemFw:
This PR add generic derivation for `Show`, with the limitation that it doesn't work for ADTs which are _both_ recursive _and_ generic in one or more type parameters.

This is due to the way Shapeless `describe` is currently implemented: since `Typeable` instances have to be able to cast, they happen to have stricter constraints than what's needed for this use case of  `describe`. I don't think the limitation in this PR can be overcome short of splitting `describe` out of `Typeable`

@kailuowang The only thing I can't understand: in my tests, `import show._` doesn't work, I have to do `import MkShow._`. Since I've never used export hook, could you take a look and see if everything is okay on that front?

### DIFF
--- a/core/src/test/scala/cats/derived/show.scala
+++ b/core/src/test/scala/cats/derived/show.scala
@@ -1,0 +1,55 @@
+package cats.derived
+
+import cats.Show
+import cats.instances.all._
+import shapeless.test.illTyped
+
+import MkShow._
+import TestDefns._
+
+class ShowTests extends KittensSuite {
+  test("Simple case classes") {
+    val foo = Foo(42, Option("Hello"))
+    val printedFoo = "Foo(i = 42, b = Some(Hello))"
+
+    assert(foo.show == printedFoo)
+  }
+
+  test("Nested case classes") {
+    val nested = Outer(Inner(3))
+    val printedNested = "Outer(in = Inner(i = 3))"
+
+    assert(nested.show == printedNested)
+  }
+
+  test("Recursive ADTs with no type parameters") {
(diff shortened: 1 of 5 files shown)


SystemFw:
@kailuowang We should be able to merge this from now, until I manage to fix: https://github.com/milessabin/shapeless/issues/750

core/src/test/scala/cats/derived/show.scala
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  test("Non recursive ADTs with type parameters") {
+    val genAdt: GenericAdt[Int] = GenericAdtCase(Some(1))
+    val printedGenAdt = "GenericAdtCase(v = Some(1))"
+
+    assert(genAdt.show == printedGenAdt)
+  }
+
+  test("Recursive ADTs with type parameters are not supported") {
```
kailuowang: Totally nitpick here, can we add the shapeless ticket link here as well?

SystemFw: No that makes sense, I'll add it tomorrow (on my phone now).

SystemFw:
@kailuowang could you have a look at the export hook thing?

kailuowang:
I am still on vacation right now (no laptops). I am actually considering proposing to @milessabin a removal of export hook from kittens 1.0 (we can consider support instance priority controlling through other means in future versions). Users will need more boilerplate code to control which instance they want in scope, but the usage would be straightforward to understand and may help adoption.

SystemFw:
Thank you! As soon as I'm back from holiday I'll start working on splitting `describe` from Typeable, which hopefully should remove the current limitation

PR #50, opened 2017-08-18
Link to typelevel cats instead of @nons fork

LukaJCB:


milessabin:
Yikes! That's old!

Well spotted ... many thanks! Merging :-)

PR #54, opened 2017-08-29
Remove export hook

kailuowang:
per discussion on gitter https://gitter.im/milessabin/kittens?at=59a57c300cef4fed2eb9c639


kailuowang:
Continued in #56

PR #59, opened 2017-09-07
Removed ConsK

### LINKED ISSUE #58: Remove ConsK
`ConsK` is not part of cats or alleycats and it was used only for the derivation of `Apply`, `Applicative` and `Monad` which are removed since 2b1cf75. I guess that makes `ConsK` kind of dead code so we can remove it too.

joroKr21:
It was a helper used for `Apply`, `Applicative` and `Monad` derivation.

Resolves #58


joroKr21:
It's from alleycats actually.

PR #63, opened 2017-10-09
New resolution

### LINKED ISSUE #61: Remove unnecessary uses of Lazy
For recursive types it seems that only `Lazy[gen.Repr]` is necessary. `Lazy[H]` and `Lazy[T]` for `H :: T` as well as `Lazy[L]` and `Lazy[R]` for `L :+: R` are redundant and we can remove them. Or at least we should remove `Lazy[T]` and `Lazy[R]` since there is no chance to see the same type in this position. The problem with `Lazy` is that it increases compile time and memory footprint. Even worse thanks to milessabin/shapeless#776 which makes it difficult to diagnose OOM errors caused by implicit resolution for large ADTs.

kailuowang:
These changes will allow semi-automatic derivation for most of the type classes, 
By semi-automatic derivation I mean you would still need to write an explicit derivation for `Foo`, e.g. `implicit val foo = cats.derive.show[Foo]`, but whatever references `Foo` has, their `Show` instances will be automatically derived if not provided in the scope. 
Here is an more detailed example that I added to the README
```scala

scala> case class Address(street: String, city: String, state: String)
scala> case class ContactInfo(phoneNumber: String, address: Address)
scala> case class People(name: String, contactInfo: ContactInfo)
scala> val mike = People("Mike", ContactInfo("202-295-3928", Address("1 Main ST", "Chicago", "IL")))
scala> import cats._,cats.implicits._

scala> //existing show instance for Address
scala> implicit val addressShow: Show[Address] = new Show[Address]{ 
          def show(a: Address) = s"${a.street}, ${a.city}, ${a.state}" 
       }  

scala> implicit val peopleShow = derive.show[People] //auto derive Show for People

scala> mike.show
res0: String = People(name = Mike, contactInfo = ContactInfo(phoneNumber = 202-295-3928, address = 1 Main ST, Chicago, IL))
```

The cost is on the kittens side - I had to duplicated code at several places due to slightly different type signature on things like `IsCCons1` or `Split`, etc.  Maybe we can improve that later.
I also removed several `Lazy` usages.  Should fix #61

### DIFF
--- a/core/src/main/scala/cats/derived/package.scala
+++ b/core/src/main/scala/cats/derived/package.scala
@@ -5,16 +5,35 @@ package cats
  * Use cats.derive to explicitly derive instance instead
  */
 package object derived {
+  @deprecated("use cats.derive.emptemptyy instead", "1.0.0-RC1")
   object empty extends MkEmptyDerivation
+
   object emptyK extends MkEmptyKDerivation
+
+  @deprecated("use cats.derive.eq instead", "1.0.0-RC1")
   object eq extends MkEqDerivation
+
   object foldable extends MkFoldableDerivation
+
+  @deprecated("use cats.derive.functor instead", "1.0.0-RC1")
   object functor extends MkFunctorDerivation
+
   object iterable extends IterableDerivationFromMkIterable
+
+  @deprecated("use cats.derive.monoid instead", "1.0.0-RC1")
   object monoid extends MkMonoidDerivation
-  object monoidK extends MkMonoidKDerivation
+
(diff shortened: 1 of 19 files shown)


core/src/main/scala/cats/derived/package.scala
```diff
@@ -5,15 +5,35 @@ package cats
  * Use cats.derive to explicitly derive instance instead
  */
 package object derived {
+  @deprecated("use cats.derive.emptemptyy instead", "1.0.0-RC1")
```
LukaJCB: I believe this should be `cats.derive.empty`?

PR #68, opened 2017-11-06
update version in readme

kailuowang:


ChristopherDavenport:
Good quick fix. What about a maven badge for simplicity?

PR #71, opened 2017-11-30
Support replicateH on HNil

ceedubs:
This changes the `ReplicateH` type class to line up a little better with the
`Sequencer` type class.

This also adds implicit `Sequencer` instances for HNil. One can sequence an
empty `List`, so I can't think of any reason that they shouldn't be able to
sequence an `HNil`. Adding this support allows one to pass zero to `replicateH`.

### DIFF
--- a/core/src/main/scala/cats/replicateH/replicateH.scala
+++ b/core/src/main/scala/cats/replicateH/replicateH.scala
@@ -1,24 +1,26 @@
 package cats.replicateH
 
+import cats.Applicative
 import cats.sequence.Sequencer
 import shapeless._
 import shapeless.ops.hlist._
 
 sealed trait ReplicateH[F[_], N, A] extends Serializable {
-  type Out
+  type LOut
+  type Out = F[LOut]
   def apply(fa: F[A]): Out
 }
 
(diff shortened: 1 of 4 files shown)


ceedubs:
@milessabin @kailuowang this is the followup PR to #70.

core/src/main/scala/cats/replicateH/replicateH.scala
```diff
@@ -1,24 +1,26 @@
 package cats.replicateH
 
+import cats.Applicative
```
kailuowang: nit: is this import used?

ceedubs: Ah good catch. I was going to add the support here instead of within `Sequencer` at first. I'll fix this.

kailuowang:
do you mind we release this in 1.0.0 (synced with Cats 1.0, I am hoping this week or early next)?

ceedubs:
@kailuowang sounds good to me. This isn't exactly battle-tested, but since it mostly delegates through to `Fill` and `Sequencer`, I can't think of any issues that it might be likely to create.

PR #73, opened 2017-12-01
Restore full auto derivation for Empty and Functor

joroKr21:
Using some new combinators from shapeless 2.3.3:
* `OrElse` to prioritize existing instances during derivation
* `Refute` to avoid derivation when an instance exists

Applied successfully to:
- [x] `Empty`
- [x] `Functor`
  
Didn't find a better way to test both semi auto and full auto derivation other than duplicate all tests.

### DIFF
--- a/core/src/main/scala/cats/derived/empty.scala
+++ b/core/src/main/scala/cats/derived/empty.scala
@@ -14,39 +14,40 @@
  * limitations under the License.
  */
 
-package cats.derived
+package cats
+package derived
 
 import alleycats.Empty
-
 import shapeless._
+import scala.annotation.implicitNotFound
 
-trait MkEmpty[T] extends Empty[T]
+
+@implicitNotFound("Could not derive an instance of Empty[${A}]")
+trait MkEmpty[A] extends Empty[A]
 
 object MkEmpty extends MkEmptyDerivation {
-  def apply[T](implicit e: MkEmpty[T]): MkEmpty[T] = e
+  def apply[A](implicit empty: MkEmpty[A]): MkEmpty[A] = empty
 }
 
-private[derived] abstract class  MkEmptyDerivation extends MkEmpty1 {
-  implicit val mkEmptyHnil: MkEmpty[HNil] =
-    new MkEmpty[HNil] {
-      def empty = HNil
-    }
+private[derived] abstract class MkEmptyDerivation {
 
-  implicit def mkEmptyHconsAvailableInstance[H, T <: HList](implicit eh: Lazy[Empty[H]], et: MkEmpty[T])
-    : MkEmpty[H :: T] = mkEmptyHcons(eh.value, et)
-}
-
-private[derived] abstract class MkEmpty1 {
-  implicit def mkEmptyHconsFurtherDerive[H, T <: HList](implicit eh: Lazy[MkEmpty[H]], et: MkEmpty[T])
-  : MkEmpty[H :: T] = mkEmptyHcons(eh.value, et)
-
-  protected def mkEmptyHcons[H, T <: HList](eh: Empty[H], et: MkEmpty[T])
-  : MkEmpty[H :: T] = new MkEmpty[H :: T] {
-    val empty = eh.empty :: et.empty
+  protected def instance[A](default: => A): MkEmpty[A] = new MkEmpty[A] {
+    def empty = default
   }
 
-  implicit def mkEmptyGeneric[T, R](implicit gen: Generic.Aux[T, R], er: Lazy[MkEmpty[R]])
-    : MkEmpty[T] = new MkEmpty[T] {
-      val empty = gen.from(er.value.empty)
-    }
+  implicit val mkEmptyHNil: MkEmpty[HNil] =
+    instance(HNil)
+
+  implicit def mkEmptyHCons[H, T <: HList](
+    implicit H: Empty[H] OrElse MkEmpty[H], T: MkEmpty[T]
+  ): MkEmpty[H :: T] = instance {
+    H.unify.empty :: T.empty
+  }
+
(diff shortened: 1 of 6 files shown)


milessabin:
There's a lot of whitespace and renaming noise here which is making it hard for me to see the real impact of the change ... would you mind minimizing the diff?

joroKr21:
@milessabin I minimized the diff. The most significant impact is reducing code duplication in cases where we want to support derivation for both nested ADTs and regular implicits forcing us to have two versions of a method with the only difference being using `Functor` vs `MkFunctor` in the signature. Of course the two versions also have to be prioritized, introducing another layer of inheritance.

Aside: I will open a separate issue to discuss code formatting and maybe a small style guide for implicit heavy programming will emerge :smile:

milessabin:
Thanks for that ... it's a lot clearer what's going on now :-)

Very interesting! A couple of things ...

First, it looks general enough that I think it probably belongs in shapeless rather than kittens.

Second, do you think you could make this work with byname implicits rather than `Lazy`?

milessabin:
Oh, also it might be interesting to compare compile times using `OrElse` vs. duplication.

joroKr21:
I agree this is a good candidate to include in shapeless.

What do you mean make it work with byname implicits? Just try it out with typelevel Scala? Or something else? scala/scala#6050 isn't even merged yet.

Regarding compilation time I'd like to know too. I don't have any tools setup for this yet though. What would you recommend?

milessabin:
I mean replace the uses of `Lazy` with byname arguments.

joroKr21:
I must have overlooked that kittens is using typelevel scala :+1:

milessabin:
Ahh ... actually, so did I. @kailuowang is that still necessary?

If you check out and build my byname implicits scala/scala PR kittens should build with it, I think.

milessabin:
That said I'll be putting together a TLS release with byname implicits included very shortly, so you might be better off waiting for that.

kailuowang:
TLS is no longer necessary for [SI-7046](https://github.com/scala/scala/pull/5284]). 
If we use by-name implicit, how will that affect downstream dependencies?

joroKr21:
I tried out byname implicits and they don't play along with `IsHCons1`. I'm getting a bunch of erros for `Functor`:
```sbt
[error] Error while emitting cats/derived/FunctorSuite$anon$macro$4679$1$anon$apply$macro$4736$1$$anon$14
[error] assertion failed: ClassBType.info not yet assigned: Lcats/derived/FunctorSuite;
```
You can see it here: https://github.com/joroKr21/kittens/tree/orelse-byname

milessabin:
What Scala version is that using? It should be a local build of my PR branch masquerading as 2.13.0-M2 with `scalaOrganization := "org.scala-lang"`.

joroKr21:
I tried with `byname-implicits-2.12.x`. 2.13 leaves unresolved dependencies.

milessabin:
OK, that's not going to be a helpful comparison. Best to wait for the next TLS release.

kailuowang:
I really like this, @milessabin shall we move `OrElse` to Shapeless and kittens wait for a Shapeless release ?

milessabin:
@kailuowang yes, I think that would make sense.

joroKr21:
I will open a PR on shapeless after I do some compile times benchmarking

joroKr21:
I used this to eyeball the overhead of `OrElse` (when deriving `Show`): https://github.com/propensive/magnolia/blob/master/benchmarks/2.12/src/kittens/show_10.scala. It's not rigorous at all but for anything more extensive it would be nice to add some infrastructure to kittens.

This is the baseline:
```
kittens/show_10:
00:26.435609   src/kittens/show_10.scala
00:15.787987   src/kittens/show_10.scala
00:12.580924   src/kittens/show_10.scala
00:12.240625   src/kittens/show_10.scala
00:12.338174   src/kittens/show_10.scala
00:12.683410   src/kittens/show_10.scala
00:13.028626   src/kittens/show_10.scala
00:10.711003   src/kittens/show_10.scala
00:10.899289   src/kittens/show_10.scala
00:10.488324   src/kittens/show_10.scala
```

Using `OrElse` instead of inheritance and duplication:
```
kittens/show_10:
00:33.097576   src/kittens/show_10.scala
00:19.824288   src/kittens/show_10.scala
00:16.151477   src/kittens/show_10.scala
00:15.280592   src/kittens/show_10.scala
00:15.428244   src/kittens/show_10.scala
00:20.711654   src/kittens/show_10.scala
00:15.938819   src/kittens/show_10.scala
00:14.994564   src/kittens/show_10.scala
00:17.752814   src/kittens/show_10.scala
00:14.651043   src/kittens/show_10.scala
```

I was able to bring it back down with these changes:
* Make coproduct tails strict (by implementing `Show[CNil]`)
* Use `Generic` instead of `LabelledGeneric` for sealed traits
```
kittens/show_10:
00:28.106574   src/kittens/show_10.scala
00:16.221340   src/kittens/show_10.scala
00:13.135186   src/kittens/show_10.scala
00:12.741140   src/kittens/show_10.scala
00:11.605901   src/kittens/show_10.scala
00:12.099189   src/kittens/show_10.scala
00:11.953956   src/kittens/show_10.scala
00:12.959893   src/kittens/show_10.scala
00:10.598033   src/kittens/show_10.scala
00:10.591219   src/kittens/show_10.scala
```

So it looks like `OrElse` adds some noticeable overhead but it's not worse than `Lazy` and `LabelledGeneric`. @milessabin the more interesting question is whether `OrElse` would be too much indirection for `-Yinduction-heuristics` (assuming we have byname implicits).

kailuowang:
Thanks so much! I will take a closer look Monday.

milessabin:
Yup, this looks good ... merge away! :-)

PR #76, opened 2017-12-13
fix Show derivation in deep hierarchy

kailuowang:


core/src/main/scala/cats/derived/show.scala
```diff
@@ -85,10 +85,10 @@ trait MkShow2 extends MkShow3 {
 
   implicit def genericDerivedShowProduct[A, R <: HList](
                                                          implicit repr: LabelledGeneric.Aux[A, R],
-                                                         t: Typeable[A],
-                                                         s: MkShow[R]): MkShow[A] = instance { a =>
-    val name = t.describe.takeWhile(_ != '[')
-    val contents = s.show(repr.to(a))
+                                                         t: Lazy[Typeable[A]],
```
joroKr21: I don't think `Typeable` needs `Lazy`.

kailuowang:
@joroKr21 thanks so much for reviewing, I addressed your comments. And I made an attempt for a new way to do semi-auto derivation. I.E. instead of duplicating the code, I just have a high priority `fromShow` method
```scala
  implicit def fromShow[T](implicit s: Show[T]): MkShow[T] = new MkShow[T] {
    override def show(t: T): String = s.show(t)
  }
```
The downside is one more method delegation.

joroKr21:
Yes, that also works, but then we give up all dreams for auto derivation. Maybe it's for the best.

kailuowang:
@joroKr21 would it? even though this `fromShow` is in a different scope from your proposed 
`implicit def derivedShow[T](mk: MkShow[T]): Show[T] = mk`?

milessabin:
Clash in the sense of ambiguity?

joroKr21:
In the sense of infinite cycle `Show -> MkShow -> Show -> ...`

kailuowang:
In conclusion this semi-auto still the best solution I can think of, as it respects existing instances. The high priority `implicit def fromShow[T](implicit s: Show[T]): MkShow[T]` removed all the duplicated code at the cost of one more method delegation.

kailuowang:
So WDYT @milessabin ?

milessabin:
LGTM.

PR #78, opened 2017-12-16
(discussion) Yet another idea for auto derivation

joroKr21:
Forgive me for throwing so many different ideas at you :smile: Here is another one:

We can emulate the priority example from scala/scala#6139 with a phantom type parameter like so:
```scala
class MkShow[+P <: Priority, A](val instance: Show[A]) extends AnyVal

trait MkShowDerivation extends MkShow0 {
  def mk[P <: Priority, A](tos: A => String): MkShow[P, A] =
    new MkShow(Show.show(tos))

  implicit def fromShow[T](implicit show: Show[T]): MkShow[Hidden, T] =
    mk(show.show)
}
```

And we can have a `Hidden` priority level that is on a separate branch from other priority levels to break the cycle `Show -> MkShow -> Show -> ...` So now we can enable auto derivation:
```scala
  object show {
    implicit def mkDerivedShow[A](
      implicit ev: Refute[Show[A]], show: MkShow[LowPriority, A]
    ): Show[A] = show.instance
  }
```

`Refute` here makes sure that even instances from companion objects are respected. Users can also provide a fallback instance that has lower priority than `Derived`:
```scala
implicit def fallbackShow[A]: MkShow[Default, A] = MkShow.mk(_.toString)
```

The priority hierarchy here is the same that we had in export-hook.

### DIFF
--- a/core/src/main/scala/cats/derive.scala
+++ b/core/src/main/scala/cats/derive.scala
@@ -1,9 +1,12 @@
 package cats
-import alleycats.{Empty, EmptyK, Pure}
+
+import alleycats._
 import cats.derived._
+import shapeless.Lazy
+
 
 object derive {
-  def functor[F[_]](implicit F: MkFunctor[F]) : Functor[F] = F
+  def functor[F[_]](implicit F: Lazy[MkFunctor.LowPriority[F]]): Functor[F] = F.value.instance
   def empty[A](implicit A: MkEmpty[A]): Empty[A] = A
   def emptyK[F[_]](implicit F: MkEmptyK[F]): EmptyK[F] = F
   def eq[T](implicit F: MkEq[T]): Eq[T] = F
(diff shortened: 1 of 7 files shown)


kailuowang:
This is really cool. This would even use the imported `Show[Option[A]]` instance right? Can we add a test? The auto derived show instance displays a `None.type()` for `None`, using this it should just display `None`. 
Would this work for HKT?

joroKr21:
Yes, the existing instance should be respected by auto derivation. With semiauto ofc it's not possible. But it seems that to make it work with HKT the scheme needs to change to using `MkFunctor[F] with P <: Priority`. The extra type parameter is confusing for implicit resolution.

kailuowang:
I like the `with P` scheme even better

joroKr21:
The specific case of implementing `MkFunctor` is more difficult since it extends `Functor` with a new method `safeMap`. I can't figure out how to implement it as a wrapper rather than a subclass, because then I can't match like it's done [here](https://github.com/milessabin/kittens/blob/master/core/src/main/scala/cats/derived/functor.scala#L114).

kailuowang:
@joroKr21 any reason you closed this?

joroKr21:
Either I missclicked or I pushed the wrong branch and there was no diff and it auto closed.

joroKr21:
`Functor` worked out, but the default `MkShow` example now doesn't work (it's ambiguous with the `Hidden` one).

joroKr21:
I am now convinced that the complexity of this scheme brings very little benefit and doesn't pull its weight. The prioritization hierarchy itself can cause ambiguities and therefore it's not suitable for users to plug in their own derivations. IMO the better option for full auto derivation is #73. Ok to close this PR?

PR #81, opened 2018-01-04
set default project to root

kailuowang:
I kept forgetting to use the `releaseAll` alias in kittens. @milessabin can we change the default project back to `root`, it would be a little cumbersome to type `jvm` every sbt session, but it's less surprising.

### DIFF
--- a/.sbtrc
+++ b/.sbtrc
@@ -1,1 +1,1 @@
-alias boot = ;reload ;project coreJVM ;iflast shell
+alias boot = ;reload ;project root ;iflast shell


milessabin:
Is there some way to make release in subprojects fail or warn instead?

kailuowang:
Hmmm maybe I can use the `releaseAll` to set a setting key and in the release step check that setting key. Would that be too convoluted?

milessabin:
TBQH, I think you should go with whatever option works best for you.

kailuowang:
Okay, I am going to merge this one then. I am so used to switch to the JVM module in some many projects.

PR #83, opened 2018-01-16
restore auto-derivation for show

kailuowang:
@joroKr21 mind to take a look? 
Once this is approved I'll add more. 
I also namespaced the implicit methods to avoid potential conflicts.

### DIFF
--- a/core/src/main/scala/cats/derived/show.scala
+++ b/core/src/main/scala/cats/derived/show.scala
@@ -49,9 +49,9 @@ trait MkShow1 extends MkShow2 {
   // used when Show[V] (a member of the coproduct) has to be derived.
   implicit def coproductDerivedShow[K <: Symbol, V, T <: Coproduct](
      implicit key: Witness.Aux[K],
-     showV: Lazy[MkShow[V]],
+     showV: Show[V] OrElse MkShow[V],
      showT: MkShow[T]): MkShow[FieldType[K, V] :+: T] = instance {
-    case Inl(l) => showV.value.show(l)
+    case Inl(l) => showV.unify.show(l)
     case Inr(r) => showT.show(r)
   }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/derived/show.scala
```diff
@@ -21,23 +21,16 @@ object MkShow extends MkShowDerivation {
...
-
-trait MkShow0 extends MkShow1 {
+trait MkShowDerivation extends MkShow1 {
   implicit val emptyProductDerivedShow: MkShow[HNil] =
     instance(_ => "")
 
   implicit def productDerivedShowFurther[K <: Symbol, V, T <: HList](
        implicit key: Witness.Aux[K],
-       showV: Lazy[MkShow[V]],
+       showV: Show[V] OrElse MkShow[V],
```
joroKr21: I think we should do this for coproducts as well. Although less likely that someone would define a special case only for one subclass, it's still a possibility.

kailuowang: thanks I missed that.

milessabin:
Looks good, but there are a couple of renamings that appear to be unrelated to the essence of the PR: `mkEmpty` -> `kittensMkEmpty` and `mkFunctor` -> `kittensMkFunctor`. Why are they included here?

kailuowang:
@milessabin they are to avoid implicit conflicts. In this PR, I introduced a new one of the same kind, namely `kittenMkShow`. To make it consistent I renamed the other two and hopefully setup a convention for naming schema. We can certainly do that in a separate PR if you prefer.

milessabin:
What do they conflict with?

kailuowang:
Nothing in particular at the moment, but if some other lib for some reason also has an implicit method called `mkShow` which isn't very inconceivable, it will conflict with our import. This is to avoid the potential implicit name conflict when both libs are imported.

milessabin:
Gotcha. OK, no objections ... LGTM :-)

kailuowang:
thanks!

Issue #82, opened 2018-01-16
Restore full auto derivation


joroKr21:
Let's divide and conquer.
Similar to #73 
If you have time, pick one and comment here:

- [x] `Empty`
- [ ] `EmptyK`
- [ ] `Eq`
- [ ] `Foldable`
- [x] `Functor`
- [ ] `Monoid`
- [ ] `MonoidK`
- [ ] `Pure`
- [ ] `Semigroup`
- [ ] `SemigroupK`
- [x] `Show`

kailuowang:
Thanks @joroKr21 
I will work on `Show` today.

joroKr21:
Sorry, I didn't have much time for this. Working on `EmptyK` now.
Duplicating the tests is a bit annoying but I don't see anything we can do here.

kailuowang:
I may find some time to work on `Eq` today

kailuowang:
fixed by #85

PR #84, opened 2018-01-31
Added missing induction step for products in Foldable

mfolnovic:
Derivation of Foldable was missing part of induction step. 

Copy & Pasted from derivation of Functor, with changes.

I haven't tested this at all (except `sbt compile`), I don't really know how to. :(


kailuowang:
Thanks. This mechanism is no longer the desired approach now that we are taking the design by @joroKr21 described in #82

kailuowang:
I am going to close this one for now. Feel free to reopen if you want to continue discussion / work

PR #85, opened 2018-01-31
restore full auto derivation for all type classes. ready for RC3

kailuowang:
also using `Cached` to improve performance when doing full auto derivation (otherwise it will be re-derived every usesite). cc @joroKr21


kailuowang:
The `Cached` is breaking the tests where instances are derived using different context. How about we provide two different version of auto derivation, one cached and one non-cached. WDYT @milessabin

milessabin:
Unfortunately I think that this is a fundamental limitation of the implementation of `Cached` in shapeless. TBQH I'd recommend against using it in a library ... if it's used at all it should be in end-user code.

I'm working on a language-level replacement, though I don't have a timetable for that just yet.

kailuowang:
@milessabin @joroKr21 so here is a new organization I'd like to propose and see what do you guys think. 

We provide three objects: `auto`, `cached` and `semi`
The recommended best practice is going to be a semi auto one:
```scala
implicit val showFoo: Show[Foo] = {
   import auto._
   semi.show
}
```
This will respect all existing instances even if the field is a type constructor. For example `Show[List[A]]` will use the native `Show` instance for `List` and derived instance for `A`. And it manually caches the result to the `val showFoo`. Downside user will need to write one for every type they directly need a `Show` instance

There are 3 alternatives:
1. full auto: 
```scala
import auto.show._
```
The downside is that it will re-derive for every use site, which multiples the compilation time cost. 

2. full auto cached 
```scala 
import cached.show._
```
This one caches the derived instance globally. So it's only applicable if the instance is global in the application. This could be problematic for libraries, which has no control over the uniqueness of an instance on use site. 

3. manual semi
```scala
implicit val showFoo: Show[Foo] =  semi.show
```
It has the same downside as the recomended semi-auto practice but also suffers from the type constructor field issue. I.e. if a field type is a type constructor whose native instance relies on the instance of the parameter type, this approach will by default derive an instance for the type constructor one. To overcome this user have to first derive the instance for type parameter. 
e.g.  given
```scala
case class Foo(bars: List[Bar])
case class Bar(a: String)
```
user will need to drive Bar first and then Foo
```scala
implicit val showBar: Show[Bar] =  semi.show
implicit val showFoo: Show[Foo] =  semi.show
```
This way the native instance for `Show[List]` would be used.

milessabin:
I'll leave it to your judgement, but bear in mind that while `Cached` works well for some people in some circumstances, it's buggy (in that it equates instances which it shouldn't) and will be going away once I've worked out a mechanism for sharing instances as part of the language.

kailuowang:
@milessabin updated to the new org and I think  ready for RC3 release.

joroKr21:
Sorry for dropping the ball here, got really busy recently. I don't mind the cached version, for people that only use coherent instances it has no downside.

PR #86, opened 2018-03-22
Update dep in README to RC3

lJoublanc:
For some reason RC2 is missing `auto` object when I try to compile - RC3 works fine though.


kailuowang:
thanks!

Issue #87, opened 2018-04-04
Deriving Order


torgeirsh:
There seems to be no support for deriving Order. Is it just an oversight, or is a technical issue preventing it? shapeless-contrib used to have support for scalaz Order.

milessabin:
Not so much an oversight as Just Missing ... do you want to take a swing at it?

torgeirsh:
I had a brief try at duplicating the code for Eq, but I just get "could not find implicit value" errors, and unfortunately can't spend time debugging and getting familiar with the internals right now.

kailuowang:
@torgeirsh if you kindly submit a PR with your WIP, I will try take a stab when I get some free time.

torgeirsh:
@kailuowang Sorry, but I didn't keep the code, as it was a very quick hack outside the project tree. Basically I just copied the code for deriving Eq to a new file, replaced everything with Order, and sprinkled NotImplementedExceptions in the compare implementations.

kailuowang:
For products with multiple fields how do we order? prioritize from left to right? 
How do we order coproduct? maybe only provide PartialOrder for coproduct?

milessabin:
I'd say yes to both.

torgeirsh:
For what it's worth, shapeless-contrib provided scalaz.Order for coproduct as well, with the following implementation:
```
   def order(x: λ, y: λ) = (x, y) match {
	    case (Inl(a), Inl(b)) => L.order(a, b)
	    case (Inl(_), Inr(_)) => Ordering.LT
	    case (Inr(_), Inl(_)) => Ordering.GT
	    case (Inr(a), Inr(b)) => R.order(a, b)
	  }
```
I don't know what the pros and cons are, compared to PartialOrder.

kailuowang:
I don't know how I feel about that. It seems even more arbitrary than the left to right order in product. For product, another question poped up: shall we skip on fields without order instance or simply fail the derivation there. 
Now I think about it, I think it's rather arbitrary for classes with more than one field, and auto-derivation is not that necessary for classes with one field. 
I am curious what's your actual use case @torgeirsh?

milessabin:
I would fail if any fields don't have instances.

kailuowang:
"when any ordering" will do how about just use the hashCode? 
```scala
implicit def arbOrder[A](implicit r: Refute[Order[A]]): Order[A] = Order.by((_:A).hashCode)
```
[...]

PR #90, opened 2018-05-28
add PartialOrder

Qi77Qi:
For https://github.com/milessabin/kittens/issues/87

Mostly copied `Eq`.

Created https://github.com/milessabin/kittens/pull/90/files#diff-bf34c2a38c2ed0d1879d7c494ef98de3R92 because seems there's no `Order` for `Map[Int, String]`

### DIFF
--- a/core/src/main/scala/cats/derived/package.scala
+++ b/core/src/main/scala/cats/derived/package.scala
@@ -98,6 +98,12 @@ object cached {
 
   }
 
+  object partialOrder {
+    implicit def kittensMkPartialOrder[A](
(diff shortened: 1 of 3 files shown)


kailuowang:
Thanks so much @Qi77Qi !  One tiny thing, to be consistent, would you add an implicit object also to the `cats.derived.cached` object, like [this `eq` one](https://github.com/Qi77Qi/kittens/blob/b9613249a36811bb831d421621bf84ec6380e601/core/src/main/scala/cats/derived/package.scala#L94-L99)?

Qi77Qi:
@kailuowang  just updated..let me know if there's anything else!

PR #91, opened 2018-05-30
add Order derivation

Qi77Qi:


core/src/test/scala/cats/derived/order.scala
```diff
@@ -0,0 +1,138 @@
...
+  {
+
+    import auto.order._
+    import cats.instances.all._
+
+    checkAll("Outer", OrderTests[Outer].order)
+  }
+
+
+  test("IList Order consistent with universal equality")(check {
```
kailuowang: `IList` still in the test description.

kailuowang:
thanks very much!

PR #92, opened 2018-05-30
fix tests

kailuowang:
The original tests were messed up. It should import the string instance so that `Option[String]` instance *could* be derived, the test then provides a "nasty" local `Option[A]` instance that gets used by the derivation. The intention is to test if the local instance bypasses the need to do derivation.


kailuowang:
sorry Miles, I decided to revamp these tests more thoroughly in another PR, I am closing this.

PR #95, opened 2018-06-25
Add ShowPretty derivation

vlovgr:
This is a proof-of-concept for a `ShowPretty` derivation, which instead of:
```
People(name = Mike, contactInfo = ContactInfo(phoneNumber = 202-295-3928, address = 1 Main ST, Chicago, IL))
```

instead generates the following.
```
People(
  name = Mike,
  contactInfo = ContactInfo(
    phoneNumber = 202-295-3928,
    address = 1 Main ST, Chicago, IL
  )
)
```

I've copied the tests for `Show` derivation and updated them for the pretty printing. Because we respect existing `Show` instances, things like `List` doesn't look nice as we cannot propagate the indentation level -- is there any way we can work around this?

### DIFF
--- a/core/src/main/scala/cats/derived/package.scala
+++ b/core/src/main/scala/cats/derived/package.scala
@@ -62,6 +62,12 @@ object auto {
     ): Show[A] = show
   }
 
+  object showPretty {
+    implicit def kittensMkShowPretty[A](
+     implicit refute: Refute[Show[A]], showPretty: MkShowPretty[A]
+    ): ShowPretty[A] = showPretty
+  }
+
   object monoid extends MkMonoidDerivation //todo the regular approach doesn't work for monoid
 
   object semigroup {
@@ -139,6 +145,12 @@ object cached {
     : Show[A] = ev.value
   }
 
+  object showPretty {
+    implicit def kittensMkShowPretty[A](
+      implicit refute: Refute[Show[A]], ev: Cached[MkShowPretty[A]])
+    : ShowPretty[A] = ev.value
+  }
+
   object monoidK {
     implicit def kittensMkMonoidK[F[_]](
       implicit refute: Refute[MonoidK[F]], ev: Cached[MkMonoidK[F]]
@@ -202,12 +214,14 @@ object semi {
   def partialOrder[A](implicit ev: MkPartialOrder[A]): PartialOrder[A] = ev
 
   def order[A](implicit ev: MkOrder[A]): Order[A] = ev
-  
+
   def hash[A](implicit ev: MkHash[A]): Hash[A] = ev
 
   def functor[F[_]](implicit ev: MkFunctor[F]): Functor[F] = ev
 
   def show[A](implicit ev: MkShow[A]): Show[A] = ev
+
+  def showPretty[A](implicit ev: MkShowPretty[A]): ShowPretty[A] = ev
 
   def foldable[F[_]](implicit F: MkFoldable[F]): Foldable[F] = F
 
--- a/core/src/main/scala/cats/derived/showPretty.scala
+++ b/core/src/main/scala/cats/derived/showPretty.scala
@@ -1,0 +1,84 @@
+package cats.derived
+
+import cats.Show
+import shapeless._
+import shapeless.labelled.FieldType
+
+trait ShowPretty[A] extends Show[A] {
+  def showWithIndent(a: A, indent: Int): String
+  def show(a: A): String = showWithIndent(a, 0)
+}
+
+object ShowPretty {
+  implicit def fromShow[A](implicit s: Show[A]): ShowPretty[A] =
+    new ShowPretty[A] {
+      override def showWithIndent(a: A, indent: Int): String =
+        s.show(a)
+    }
+}
+
+trait MkShowPretty[A] extends ShowPretty[A]
+
+object MkShowPretty extends MkShowPrettyDerivation {
+  def apply[A](implicit showPretty: MkShowPretty[A]): MkShowPretty[A] =
+    showPretty
+}
+
+trait MkShowPrettyDerivation extends MkShowPretty1 {
+  implicit val emptyProductDerivedShowPretty: MkShowPretty[HNil] =
+    instance((_, _) => "")
+
+  implicit def productDerivedShowPretty[K <: Symbol, V, T <: HList](
+      implicit key: Witness.Aux[K],
+      showV: ShowPretty[V] OrElse MkShowPretty[V],
+      showT: MkShowPretty[T]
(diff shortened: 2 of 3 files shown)


kailuowang:
Thanks very much @vlovgr. One alternative might be instead of propagate the indent down, we could let the parent do the indentation for children? That might means that parent need to split children string into multiple lines and add the indentation. So it might be advantages to change ShowPretty typeclass to reaturn a list of strings each representing 1 line and then congregate at the last `show`?

vlovgr:
@kailuowang I did a quick test of the approach you suggested. I think the implementation is not as straightforward and (unless I'm missing something) also can't workaround the issue with having to respect existing `Show` instances for e.g. `List`.

Here's the quick implementation which also passes the current tests without changes.

```scala
package cats.derived

import cats.Show
import shapeless._
import shapeless.labelled._

trait ShowPretty[A] extends Show[A] {
  def showLines(a: A): List[String]
  def show(a: A): String = showLines(a).mkString("\n")
}

object ShowPretty {
  implicit def fromShow[A](implicit s: Show[A]): ShowPretty[A] =
    new ShowPretty[A] {
      override def showLines(a: A): List[String] =
        List(s.show(a))
    }
}

trait MkShowPretty[A] extends ShowPretty[A]

object MkShowPretty extends MkShowPrettyDerivation {
  def apply[A](implicit showPretty: MkShowPretty[A]): MkShowPretty[A] =
    showPretty
}

trait MkShowPrettyDerivation extends MkShowPretty1 {
  implicit val emptyProductDerivedShowPretty: MkShowPretty[HNil] =
    instance(_ => Nil)

  implicit def productDerivedShowPretty[K <: Symbol, V, T <: HList](
      implicit key: Witness.Aux[K],
      showV: ShowPretty[V] OrElse MkShowPretty[V],
      showT: MkShowPretty[T]
  ): MkShowPretty[FieldType[K, V] :: T] = instance { fields =>
    val fieldName = key.value.name
    val fieldValueLines = showV.unify.showLines(fields.head)
    val nextFields = showT.showLines(fields.tail)

    val fieldValue = {
      val head = fieldValueLines.headOption.mkString
      if (nextFields.isEmpty || fieldValueLines.size > 1) head
      else s"$head,"
    }

    val remainingLines =
      if (fieldValueLines.size > 1) {
        val tail = fieldValueLines.tail
        if (nextFields.isEmpty) tail
        else tail.init ++ tail.lastOption.map(s => s"$s,")
      } else Nil

    List(s"$fieldName = $fieldValue") ++ remainingLines ++ nextFields
  }

  implicit def emptyCoproductDerivedShowPretty: MkShowPretty[CNil] =
    instance(_ => Nil)
}

trait MkShowPretty1 extends MkShowPretty2 {
  implicit def coproductDerivedShowPretty[K <: Symbol, V, T <: Coproduct](
      implicit key: Witness.Aux[K],
      showV: ShowPretty[V] OrElse MkShowPretty[V],
      showT: MkShowPretty[T]
  ): MkShowPretty[FieldType[K, V] :+: T] = instance {
    case Inl(l) => showV.unify.showLines(l)
    case Inr(r) => showT.showLines(r)
  }
}

trait MkShowPretty2 extends MkShowPretty3 {
  implicit def genericDerivedShowPrettyProduct[A, R <: HList](
      implicit repr: LabelledGeneric.Aux[A, R],
      t: Typeable[A],
      s: Lazy[MkShowPretty[R]]
  ): MkShowPretty[A] = instance { a =>
    val name = t.describe.takeWhile(_ != '[')
    val contentLines = s.value.showLines(repr.to(a))
    val contents = contentLines.map(s => s"  $s")
    List(s"$name(") ++ contents ++ List(")")
  }
}

trait MkShowPretty3 {
  protected def instance[A](f: A => List[String]): MkShowPretty[A] =
    new MkShowPretty[A] {
      def showLines(a: A): List[String] = f(a)
    }

  implicit def genericDerivedShowPrettyCoproduct[A, R <: Coproduct](
      implicit repr: LabelledGeneric.Aux[A, R],
      s: Lazy[MkShowPretty[R]]
  ): MkShowPretty[A] =
    instance(a => s.value.showLines(repr.to(a)))
}
```

kailuowang:
@vlovgr thanks so much for trying it. what is the result for a list with this implementation? I guess I probably didn't fully understand your indentation problem. I thought you are referring that the list string resets the indentation (everything inside goes back to leftmost). With this implementation at least the content inside list will honor the inherited indentation, no? 
`Foo(a = "a", l = List(Bar(b = "b")))` under the new implementation would look like 
```
Foo(
  a = a
  l = List(Bar(
    b = "b"
  ))
)
```

No?

vlovgr:
@kailuowang In this particular case, the issue is that cats provides a `Show[List[A]]` and when we generate the `Show` (`ShowPretty`) for `A`, there's no way to pass the indentation to `ShowPretty[A]` via `Show[List[A]]`. In general, I'm not that keen to go down the route of modifying results from arbitrary `Show` instances, so not sure there's much we can do here? We still want to respect the existing instances. WDYT?

kailuowang:
@vlovgr the `ShowPretty[A]` only need to care about its own indentation inside its string block. The `Show[List[A]]` is going to simply take that string in. Even if we want to generate a `ShowPretty` instance for types like `List`,  I don't see how it's possible without a deeper understanding/analysis of its data structure. It seems to me that the best thing we can do is to let the parent `ShowPretty[Foo]` simply indent the string from `Show[List[A]]` and to do that it needs to add some prefix space to each line. and if inner ShowPertty[A] does it's own indentation, we will have the list indentation "inconsistent" at its level, but at least the rest is consistent right?  
Basically if we let parent take care of indentation we got the following, where the list is not indenting, but overrall we only got one indentation wrong. 
```
Foo(
  a = a
  l = List(Bar(
    b = "b",
    c = "c",
  ))
)
```
v.s.  let the Show[List] reset the indentation to 0
```
Foo(
  a = a
  l = List(Bar(
  b = "b",
  c = "c",
))
)
```

vlovgr:
@kailuowang Yes, I agree that it's better that way. Just like you proposed, it's enough to change to:
```scala
val fieldValueLines = showV.unify.showLines(fields.head).flatMap(_.split("\\n"))
```
in the quick implementation above to get the desired behaviour. I've updated the pull request with these changes. I'm not entirely happy with the code (it was put together quickly) but won't have time to clean it up for a couple of weeks. If you want, feel free to make any changes you think are sensible before then. Thanks a lot for all the help!

kailuowang:
@vlovgr thanks very much. I made one small change in your PR. moving the split to where `ShowPretty` derived from `Show`. This way we only need to do the split once.

PR #103, opened 2018-08-20
Implement MkTraverse

letalvoj:
I wanned to learn how the type-class derivation work. Thus I attempted to implement derivation for `Traverse`, since it was not present in kittens.

After a brief discussion on gitter I added it to kittens as a separate class with the _fold_ methods implemented using `traverse`.

## Further TODOs

 * [x] write unit tests
 * [x] make it stacksafe
 * [x] attribute the [original author](https://github.com/typelevel/cats/issues/107#issuecomment-393797529) of default implementation of fold methods

## Questions

 * Is the code style fine?
 * Should all the files have a licence header?

### DIFF
--- a/core/src/main/scala/cats/derived/package.scala
+++ b/core/src/main/scala/cats/derived/package.scala
@@ -140,6 +140,18 @@ object cached {
     : Functor[F] = ev.value
   }
 
+  object foldable {
+    implicit def kittensMkFoldable[F[_]](
+       implicit refute: Refute[Foldable[F]], ev: Cached[MkFoldable[F]])
+    : Foldable[F] = ev.value
+  }
+
+  object traverse{
+    implicit def kittensMkTraverse[F[_]](
+       implicit refute: Refute[Traverse[F]], ev: Cached[MkTraverse[F]])
+    : Traverse[F] = ev.value
+  }
+
   object show {
     implicit def kittensMkshow[A](
       implicit refute: Refute[Show[A]], ev: Cached[MkShow[A]])
@@ -225,6 +237,8 @@ object semi {
   def showPretty[A](implicit ev: MkShowPretty[A]): ShowPretty[A] = ev
 
   def foldable[F[_]](implicit F: MkFoldable[F]): Foldable[F] = F
+
+  def traverse[F[_]](implicit F: MkTraverse[F]): Traverse[F] = F
 
   def monoid[T](implicit T: MkMonoid[T]): Monoid[T] = T
 
--- a/core/src/main/scala/cats/derived/package.scala
+++ b/core/src/main/scala/cats/derived/package.scala
@@ -92,6 +92,7 @@ object auto {
 
   object foldable extends MkFoldableDerivation
 
+  object traverse extends MkTraverseDerivation
 
 }
 
--- a/core/src/main/scala/cats/derived/traverse.scala
+++ b/core/src/main/scala/cats/derived/traverse.scala
@@ -2,7 +2,7 @@ package cats.derived
 
 import cats.derived.MkTraverse.SafeTraverse
 import cats.syntax.all._
-import cats.{Applicative, Eval, Monoid, Traverse}
+import cats.{Applicative, Endo, Eval, Monoid, Traverse}
 import shapeless._
 
 import scala.annotation.implicitNotFound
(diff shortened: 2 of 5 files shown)


letalvoj:
Ok, I have an issue. I'll try to dig deeper into it.

This works
```scala
scala> case class Mine[T](u: T, v: T, i: String, j: String)
scala> println(MkTraverse[Mine].traverse(Mine[String]("12","ab","_","!"))(_.toCharArray.toList))
List(Mine(1,a,_,!), Mine(1,b,_,!), Mine(2,a,_,!), Mine(2,b,_,!))
```

This fails
```scala
scala> case class Mine[T](u: T, i: String, v: T, j: String)
scala> println(MkTraverse[Mine].traverse(Mine[String]("12","ab","_","!"))(_.toCharArray.toList))
scala> <console>:25: error: Could not derive an instance of Traverse[Mine]
       println(MkTraverse[Mine].traverse(Mine[String]("12","ab","_","!"))(_.toCharArray.toList))
```

letalvoj:
Inspired by the `Functor` derivation I solved the problem mentioned above.

joroKr21:
Nice, I always thought kittens already had `Traverse` but it has something else instead: [traverse.scala](https://github.com/typelevel/kittens/blob/master/core/src/main/scala/cats/sequence/traverse.scala)

For testing you should be able to just check the `Traverse` laws, right?

letalvoj:
@joroKr21 I'll use the `cats.laws` module 👍

Btw. how do you run anything from this project from InteliJ? SBT works just fine from me, but the IDE does not see any classes on the classpath and ignores the kind-projector plugin 🙄

```scala
Unable to load a Suite class. This could be due to an error in your runpath. Missing class: cats.derived.TraverseSuite
java.lang.ClassNotFoundException: cats.derived.TraverseSuite
```

letalvoj:
Since `functor` has a `safeMap` method, I kind of feel like the derived `traverse` method should be _safe_ aswell. Should I implement it in terms of

```scala
def safeTraverse[G[_] : Applicative, A, B](fa: F[A])(f: A => Eval[G[B]]): Eval[G[F[B]]]
```

?

core/src/main/scala/cats/derived/package.scala
```diff
@@ -139,6 +140,18 @@ object cached {
     : Functor[F] = ev.value
   }
 
+  object foldable {
+    implicit def kittensMkFoldable[F[_]](
+       implicit refute: Refute[Foldable[F]], ev: Cached[MkFoldable[F]])
```
letalvoj: I hope that it was not missing on purpose.

core/src/main/scala/cats/derived/traverse.scala
```diff
@@ -0,0 +1,121 @@
...
+}
+
+trait MkTraverse0 extends MkTraverse1 {
+  // Induction step for products
+  implicit def mkTraverseHcons[F[_]](implicit ihc: IsHCons1[F, TraverseOrMk, MkTraverse]): MkTraverse[F] =
+    new MkTraverse[F] {
+      override def safeTraverse[G[_] : Applicative, A, B](fa: F[A])(f: A => Eval[G[B]]): Eval[G[F[B]]] = {
+        for {
+          ht <- Eval.now(ihc.unpack(fa))
+          th <- ihc.fh.unify.safeTraverse(ht._1)(f)
```
letalvoj: Curiously, when implemented with `Applicative[Eval].compose[G]`, it was not stacksafe... The `Eval.flatMap` was necesseary. I need to dig deeper to know why.

core/src/test/scala/cats/sequence/TraverseHListSuite.scala
```diff
@@ -11,7 +11,7 @@ import cats.derived._
 import org.scalacheck.Prop.forAll
 
 
-class TraverseSuite extends KittensSuite {
+class TraverseHListSuite extends KittensSuite {
```
letalvoj: Is the name fine? I needed to disambiguate the test suites and IMO this one was ill named.

letalvoj:
Oh... I'll fix the *Scala 2.13* build sometime today 🤔

milessabin:
This looks very nice ... thanks!

@kailuowang is the primary maintainer these days, so I'll defer to him.

letalvoj:
@milessabin thanks ;-)

I kind of feel like the issue which I had with the 2.13 build affects `foldable` as well, it's not just covered by tests. I'll verify that and if so I'll open yet another quick PR.

letalvoj:
Well I wrote some tests to cover the default methods and the `foldRight` is not stacksafe :(
https://github.com/typelevel/cats/issues/107#issuecomment-417694085

I'll reuse the current impl of `foldable` then ;-)

letalvoj:
Ok I did not know how to preserve stack-safety with the default fold implementations. Thus I reused the guts from `foldable`.

core/src/test/scala/cats/derived/traverse.scala
```diff
@@ -0,0 +1,163 @@
...
+package derived
+
+import cats.derived.TestDefns._
+import cats.implicits._
+import cats.laws.discipline.{FoldableTests, TraverseTests}
+import org.scalacheck.Test
+import org.scalatest.FreeSpec
+import shapeless.test.illTyped
+
+class TraverseSuite extends FreeSpec {
```
kailuowang: Nitpick: is there any technical reason that this test uses `FreeSpec` vs. the `KittenSuite` based of `FunSpec`?

letalvoj: I actually copy pasted the `functor` test suite and adapted it to `traverse`

PR #104, opened 2018-08-24
Derivation of half of the type-classes fails on simple case class

letalvoj:
# Summary
In Scala 2.13 I mentioned that there was a problem with implicit defs for both HCons and CCons being applicable on my simple `case class` with generic params interleaved with primitive values.

So far this PR adds failing tests to the test suites. Which broke more than I expected. For example the error log from `foldabe` goes crazy.

A fix could be found for example in `functor`, which is unaffected by the original problem. It exposes only `new Mk*[Const[T]#λ]`, keeping the remaining definitions `private[derived]`.

# Question

I just need to verify that this actually is a problem of kittens, not shapeless. Since, if I underestand it correctly it does not make sense to have both `HCons` and `CCons` applicable.

# TODOs

 * [x] add failing unit tests
 * [ ] fix by making some implicit defs private

### DIFF
--- a/core/src/test/scala/cats/derived/adtdefns.scala
+++ b/core/src/test/scala/cats/derived/adtdefns.scala
@@ -21,6 +21,9 @@ import org.scalacheck.{Cogen, Arbitrary}, Arbitrary.arbitrary
 import scala.annotation.tailrec
 
 object TestDefns {
+
+  case class Interleaved[T](i: Int, t: T, d: Double, tt: List[T], s: String)
+
   sealed trait IList[A]
   final case class ICons[A](head: A, tail: IList[A]) extends IList[A]
   final case class INil[A]() extends IList[A]
--- a/core/src/test/scala/cats/derived/empty.scala
+++ b/core/src/test/scala/cats/derived/empty.scala
@@ -43,7 +43,7 @@ class EmptySuite extends FreeSpec {
     }
 
     "derives an instance for Interleaved[T]" in {
-      assertCompiles("semi.empty[TestDefns.Interleaved[Int]]")
+      semi.empty[TestDefns.Interleaved[Int]]
     }
 
   }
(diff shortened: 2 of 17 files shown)


kailuowang:
Interesting. Thanks for reporting. Let me know if you intend to try fix (package private) you are proposing. I think that's probably the right / better design anyway.

letalvoj:
You mean something more specific than the failing tests which I already
pushed?

milessabin:
I mean could you see if you can put together a failing test against shapeless itself :-)

milessabin:
It'd be a failure to find an instance of `Generic1` for `Interleaved`.

letalvoj:
@kailuowang I'll give it a shot this week ;-)
@milessabin 👍I'll see

No commitment to a particular day though. We have a newborn at home. I actually implemented a first version of the `traverse` derivation when I was waiting in a maternity hospital 😅

kailuowang:
Congratulations! That's way more important and exciting than any of this. Enjoy your time with your baby.

joroKr21:
It looks like there a lot of imports missing :thinking:

joroKr21:
Also some instances don't work:
  * `EmptyK[Interleaved]`: It holds a reference to a `T`, hence it cannot be empty
  * `SemigroupK[Interleaved]`: It holds a reference to `T` which we don't know how to combine
  * `MonoidK[Interleaved]`: Well this is just `EmptyK + SemigroupK`
  * `Pure`: It holds references to primitive values - we're not using `Empty` to derive `Pure` although we could (should we?)

joroKr21:
@letalvoj would you mind if I cherry-pick your commits on a new branch from master (this one is quite old) and submit a new PR with the fixed valid tests and deleted invalid ones?

letalvoj:
NP. Go for it. I do not have the time to fix it now anyway. Too many other
responsibilities.

joroKr21:
#123 adds the tests which are expected to work and fixes the 2.13 errors.

PR #105, opened 2018-08-26
Fix libraryDependencies version in readme.

mikla:


kailuowang:
Thanks!
sorry about the delay, I've been out of town.

PR #123, opened 2019-03-15
Add Interleaved tests

joroKr21:
Some instances are not valid:
  - `EmptyK`: It holds a reference to a `T`, hence it cannot be empty
  - `SemigroupK`: It holds a reference to `T` which we don't know how to combine
  - `MonoidK`: This is just `EmptyK + SemigroupK`
  - `Pure`: It holds references to primitive values

Cherry-picked the first two commits from #104 

Supersedes #104

### DIFF
--- a/core/src/main/scala/cats/derived/foldable.scala
+++ b/core/src/main/scala/cats/derived/foldable.scala
@@ -21,9 +21,7 @@ import shapeless._
 
 trait MkFoldable[F[_]] extends Foldable[F] {
   def foldLeft[A, B](fa: F[A], b: B)(f: (B, A) => B): B = safeFoldLeft(fa, b){ (b, a) => now(f(b, a)) }.value
-
   def foldRight[A, B](fa: F[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B]
-
   def safeFoldLeft[A, B](fa: F[A], b: B)(f: (B, A) => Eval[B]): Eval[B]
 }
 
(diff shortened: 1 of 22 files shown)


joroKr21:
Ok now I see the actual 2.13 failures. Maybe related to some bug fixes around HKTs. Will take a look during the weekend.

joroKr21:
The problem is in implicit prioritization.

`mkFoldableConstFoldable`'s type signature is more specific than `mkFoldableGeneric` but it's in a super trait. `Functor` works, because `mkFunctorConst` is in a sub-trait. The fix would be to move the method.

@kailuowang I noticed that we haven't been consistent with the ordering of the implicits in different files. Also we don't use `private[derive]` everywhere and we don't use `OrElse` everywhere which means that for some typeclasses implicits from the companion objects would work and for others not. Unfortunately fixing all these problems would not be binary compatible so it would have to wait for version `2.0.0`. Do you think `kittens` major versions should keep cadence with `cats`?

joroKr21:
Hmm, it looks like we don't have mima enabled in the build. I guess we should?

kailuowang:
Thanks so much for tackling these issues. I don't see much need for kittens to sync with cats on major versions. We can go 2.0 first. BTW, cats next scheduled release is also 2.0 (but it remains BC with 1.x on core non test modules)

joroKr21:
@kailuowang the last commit fixes the `Const` implicits issue in a binary-compatible way.

I'm not sure if I should try to squash all commits? On one hand, I don't like having commits that don't compile, but on the other hand I don't know how to squash multi-author commits.

kailuowang:
@joroKr21, nice! Thanks so much! I propose we don't squash the commits in the PR, instead, when we merge PR we squash merge to master. That way the history on master are still clean without breaking commits and the PR itself preserves the history, which is referred in the merge commit.

joroKr21:
> I propose we don't squash the commits in the PR, instead, when we merge PR we squash merge to master.

Ok I will do that. But which strategy for merging PRs do we use in general? `merge`, `rebase` or `squash`?

PR #124, opened 2019-03-17
Add mima for binary compatibility check

joroKr21:
We are past version 1.0.0 so we should do this.


build.sbt
```diff
@@ -44,7 +44,8 @@ lazy val commonSettings = Seq(
     Some(ScmInfo(
       url("https://github.com/typelevel/kittens"),
       "scm:git:git@github.com:typelevel/kittens.git"
-    ))
+    )),
+  mimaPreviousArtifacts := Set("org.typelevel" %% "kittens" % "1.2.1")
```
joroKr21: 1.2.1 is the only version released for Scala 2.13

kailuowang:
Thanks!

PR #128, opened 2019-04-16
this remove the dependencies on export hook from alleycats

kailuowang:
this breaks BC. we can bump version to 2.0 but we still have some derivations whose auto don't respect existing instances
namely 
```
  object pure extends MkPureDerivation
  object foldable extends MkFoldableDerivation
  object monoid extends MkMonoidDerivation
```
Fixes for them probably will break BC too. I can't quickly figure out a way to fix them but on the other hand, it would be awkward if we have to bump to 3.0 once they are fixed. I don't see a great solution in terms of versioning here. 
For now, I vote for bump the version to 2.0.0-SNAPSHOT and disable mima before the 2.0 release.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -77,7 +77,7 @@ lazy val core = crossProject(JSPlatform, JVMPlatform).crossType(CrossType.Pure)
 lazy val coreJVM = core.jvm
 lazy val coreJS = core.js
 
-addCommandAlias("validate", ";root;clean;test;mima")
+addCommandAlias("validate", ";root;clean;test") //todo: add mima back once we release 2.0
 addCommandAlias("releaseAll", ";root;release")
 addCommandAlias("js", ";project coreJS")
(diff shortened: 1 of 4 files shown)


kailuowang:
this breaks BC. we can bump version to 2.0 but we still have some derivations whose auto don't respect existing instances
namely 
```
  object pure extends MkPureDerivation
  object foldable extends MkFoldableDerivation
  object monoid extends MkMonoidDerivation
```
Fixes for them probably will break BC too. I can't quickly figure out a way to fix them but on the other hand, it would be awkward if we have to bump to 3.0 once they are fixed. I don't see a great solution in terms of versioning here. 
For now, I vote for bump the version to 2.0.0-SNAPSHOT and disable mima before the 2.0 release.

milessabin:
I agree with @joroKr21 :+1:

PR #147, opened 2019-05-04
Fix issues with Empty derivation for coproducts

joroKr21:
When deriving an `Empty` instance for coproducts we look for a
unique member of the coproduct with an `Empty` instance.

It turns out we should use `Refute` for both the left and right sides.
E.g. `IList[Int]` shouldn't work because both `Int :: HNil` and `HNil`
have `Empty` instances.


joroKr21:
That may not be good enough either :thinking: 

I think the only way is to work with the entire `Coproduct`

joroKr21:
I haven't found a way to make it work for `Generic1` though.

PR #157, opened 2019-06-04
Scala 2.13.0 and Hash derivation rework

joroKr21:
Case classes' `hashCode` is not backwards compatible in Scala 2.13,
because it mixes the `productPrefix` for better distribution.
This means we need version specific code to pass the `Hash` laws.

  * Update dependencies for Scala 2.13.0
  * Rework `Hash` derivation for consistency
  * Split product and coproduct derivation
  * Add a Scala version specific seed for `scala.Product` subtypes
  * Optimize `HashBuilder` to a one-pass hash
  * Replace usages of `Statics` (not recommended) with `MurmurHash3`
  * Test all variants of `Hash` derivation

### DIFF
--- a/core/src/main/scala/cats/derived/hash.scala
+++ b/core/src/main/scala/cats/derived/hash.scala
@@ -1,91 +1,102 @@
-package cats.derived
+package cats
+package derived
 
-import cats.Hash
-import shapeless._, labelled._
+import scala.annotation.{implicitNotFound, tailrec}
+import scala.util.hashing.MurmurHash3
 
+@implicitNotFound("Could not derive an instance of Hash[${A}]")
 trait MkHash[A] extends Hash[A]
 
-
-
-object MkHash extends MkHash0 {
-  def apply[A](implicit hash: MkHash[A]): MkHash[A] = hash
-
-
+object MkHash extends MkHashDerivation {
+  def apply[A](implicit ev: MkHash[A]): MkHash[A] = ev
 }
 
 private[derived] trait HashBuilder[A] {
-  def hashes(a: A): List[Int]
+  def hashes(x: A): List[Int]
   def eqv(x: A, y: A): Boolean
-}
 
-object HashBuilder {
-  implicit val emptyProductDerivedHash: HashBuilder[HNil] = new HashBuilder[HNil] {
-    override def hashes(x: HNil): List[Int] = Nil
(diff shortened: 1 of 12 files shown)


core/src/main/scala/cats/derived/hash.scala
```diff
@@ -1,91 +1,91 @@
...
-    def hash(value: L :+: R): Int = value match {
-      case Inl(l) => hashV.unify.hash(l)
-      case Inr(r) => hashT.hash(r)
+private[derived] abstract class MkHashGeneric {
+  import shapeless._
+
+  implicit def mkHashGenericProduct[A, R <: HList](
+    implicit A: Generic.Aux[A, R], R: Lazy[HashBuilder[R]], ev: A <:< Product = null
+  ): MkHash[A] = instance(
+    x => R.value.hash(A.to(x), if (ev == null) MurmurHash3.productSeed else util.VersionSpecific.productSeed(x)),
```
joroKr21: I'm not sure about this to be honest.

joroKr21:
This uncovered an issue with `Hash[Long]`: typelevel/cats#2878

core/src/main/scala/cats/derived/hash.scala
kailuowang: Could we use a pattern match here? like 
```scala
x match {
   case p : Product => util.VersionSpecific.productSeed(p)
   case _ => MurmurHash3.productSeed
}
```

joroKr21: Yes, that's the part I'm not sure about. On one hand, `ev: A <:< Product = null` looks like a hack, but on the other pattern matching on the type parameter would break parametricity.

But what if the solution is to drop this law altogether from `cats-testkit`?

kailuowang: I need to think about the law. Could we split this one method into two derivation methods with different priorities? One for Product another for the rest.

joroKr21: Sure, that's also an option :thumbsup:

joroKr21: Done. We can later remove the version specific code if we decide not to enforce this law.

joroKr21:
Updated to Scala 2.13.0 :tada:

PR #161, opened 2019-06-13
Use lazy implicits in Scala 2.13

joroKr21:
Introduce version specific `Lazy` and `OrElse` which are imlpemented
with lazy implicits in Scala 2.13 and delegate to shapeless otherwise.

Unfortunately `illTyped` tests now typecheck but crash the compiler.
Looks like a compiler bug but needs investigation.

### DIFF
--- a/core/src/main/scala-2.13/cats/derived/util/versionspecific.scala
+++ b/core/src/main/scala-2.13/cats/derived/util/versionspecific.scala
@@ -3,5 +3,36 @@ package cats.derived.util
 import scala.util.hashing.MurmurHash3
 
 private[derived] object VersionSpecific {
-  def productSeed(x: Product): Int = MurmurHash3.mix(MurmurHash3.productSeed, x.productPrefix.hashCode)
+
+  def productSeed(x: Product): Int =
+    MurmurHash3.mix(MurmurHash3.productSeed, x.productPrefix.hashCode)
+
+  abstract class Lazy[+A] {
+    def value(): A
+  }
+
+  object Lazy {
+    implicit def instance[A](implicit ev: => A): Lazy[A] = () => ev
+  }
+
+  sealed trait OrElse[+A, +B] {
+    def fold[C](prim: A => C, sec: B => C): C
+    def unify[C >: A](implicit ev: B <:< C): C = fold(identity, ev)
+  }
+
+  final class Primary[+A](value: A) extends OrElse[A, Nothing] {
+    def fold[C](prim: A => C, sec: Nothing => C): C = prim(value)
+  }
+
+  final class Secondary[+B](value: => B) extends OrElse[Nothing, B] {
+    def fold[C](prim: Nothing => C, sec: B => C): C = sec(value)
+  }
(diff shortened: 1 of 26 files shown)


core/src/test/scala/cats/derived/empty.scala
```diff
@@ -52,32 +52,37 @@ class EmptySuite extends KittensSuite {
   {
     import auto.empty._
     testEmpty("auto")
-    illTyped("Empty[IList[Int]]")
-    illTyped("Empty[Snoc[Int]]")
+    // NOTE: These typecheck but crash the compiler on Scala 2.13
+//    println(Empty[IList[Int]])
+//    println(Empty[Snoc[Int]])
```
joroKr21: cc @milessabin I don't think I will be able to minimize this enough to open a Scala issue.

kailuowang:
Just curious what change did you make to pass the illtype check?

joroKr21:
> Just curious what change did you make to pass the illtype check?

That's the thing though, they are not passing. At least not for recursive types. There is one which still passes for `Empty[Rgb]` (which is not recursive and hence no intersection with lazy implicits). 

If we uncomment the `println` statements which have replaced the `illTyped` checks we can observe the compiler crash (although at a later stage).

core/src/test/scala/cats/derived/empty.scala
milessabin: Can you post the stacktrace as a gist?

joroKr21: Sure: https://gist.github.com/joroKr21/a62574d04b734628f52c0c6d4a89c7d3
There are two files: `jvm.txt` and `js.txt`, because the traces are different.

core/src/main/scala-2.13/cats/derived/util/versionspecific.scala
```diff
@@ -3,5 +3,36 @@ package cats.derived.util
...
+  final class Secondary[+B](value: => B) extends OrElse[Nothing, B] {
+    def fold[C](prim: Nothing => C, sec: B => C): C = sec(value)
+  }
+
+  object OrElse extends OrElse0 {
+    implicit def primary[A, B](implicit a: A): A OrElse B = new Primary(a)
+  }
+
+  private[util] abstract class OrElse0 {
+    implicit def secondary[A, B](implicit b: => B): A OrElse B = new Secondary(b)
```
kailuowang: @joroKr21  Sorry I missed this last time. but now I am curious why `b` here is by name (or `Lazy` in 2.12- code)?

joroKr21: Because usually it follows the pattern `UserDefinedInstance OrElse DerivedInstance` and `DerivedInstance` is usually recursive. Also it has a nice analogy to `Option.orElse` ;)

kailuowang: I see. thanks for the explanation.

PR #166, opened 2019-06-26
Ensure all derived typeclasses are Serializable

### LINKED ISSUE #130: Most derived instances are not Serializable
There is a `SerializableTest` in `KittensSuite` but it's rarely used. Unfortunately most instances cannot be made `Serializable` at the moment because of Shapeless `OrElse`.

joroKr21:
We use our own `OrElse` which we anyway do now for Scala 2.13.
This unblocks us from waiting on a shapeless release for Serializable.

Fixes #130

### DIFF
--- a/core/src/test/scala/cats/derived/show.scala
+++ b/core/src/test/scala/cats/derived/show.scala
@@ -2,14 +2,11 @@ package cats
 package derived
 
 import cats.instances.all._
+import cats.laws.discipline.SerializableTests
 
 class ShowSuite extends KittensSuite {
   import ShowSuite._
   import TestDefns._
-
-  implicit val showAddress: Show[Address] = Show.show { a =>
-    List(a.street, a.city, a.state).mkString(" ")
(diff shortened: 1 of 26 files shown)


core/src/test/scala/cats/derived/show.scala
```diff
@@ -84,9 +83,25 @@ class ShowSuite extends KittensSuite {
...
+    List(a.street, a.city, a.state).mkString(" ")
+  }
+
+  final case class Bogus(value: Int)
+  object Bogus {
+    implicit val show: Show[Bogus] = Show.show(_ => "Blah")
+  }
 
-  object semiTests {
+  object semiInstances {
```
joroKr21: I'm not sure why I had to move these instances to make them serializable. Looks like somehow `this` was captured in a closure (maybe by `Typeable` because only `Show` and `ShowPretty` are affected, and that on Scala 2.12).

PR #168, opened 2019-06-28
Derive `Contravariant`

stephen-lazaro:
First time contributing, don't usually do this kind of Shapeless-y stuff but seemed fun. Let me know if I've missed anything obvious!
Resolves some of https://github.com/typelevel/kittens/issues/144

### DIFF
--- a/core/src/main/scala/cats/derived/contravariant.scala
+++ b/core/src/main/scala/cats/derived/contravariant.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2015 Miles Sabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package cats
 package derived
 
@@ -11,10 +27,10 @@ trait MkContravariant[F[_]] extends Contravariant[F] {
   def safeContramap[A, B](fa: F[A])(f: B => Eval[A]): Eval[F[B]]
   def contramap[A, B](fa: F[A])(f: B => A): F[B] = safeContramap(fa)(f andThen Eval.now).value
 }
-object MkContravariantvariant extends MkContravariantvariantDerivation {
+object MkContravariant extends MkContravariantDerivation {
   def apply[F[_]](implicit F: MkContravariant[F]): MkContravariant[F] = F
 }
-private[derived] abstract class MkContravariantvariantDerivation extends MkContravariantvariantNested {
+private[derived] abstract class MkContravariantDerivation extends MkContravariantNested {
   implicit val mkContraHNil: MkContravariant[Const[HNil]#λ] = mkContraConst
   implicit val mkContraCNil: MkContravariant[Const[CNil]#λ] = mkContraConst
 
@@ -24,27 +40,19 @@ private[derived] abstract class MkContravariantvariantDerivation extends MkContr
     }
 }
 
-private[derived] abstract class MkContravariantvariantNested extends MkContravariantCons {
+private[derived] abstract class MkContravariantNested extends MkContravariantCons {
+  protected type FunctorOrMk[F[_]] = Functor[F] OrElse MkFunctor[F]
+  protected def mkSafeMap[F[_], A, B](F: FunctorOrMk[F])(fa: F[A])(f: A => Eval[B]): Eval[F[B]] =
+    F.unify match {
+      case mk: MkFunctor[F] => mk.safeMap(fa)(f)
+      case p => Eval.later(p.map(fa)(f(_).value))
+    }
+
   implicit def mkContravariantFunctorNested[F[_]](implicit F: Split1[F, FunctorOrMk, ContraOrMk]): MkContravariant[F] =
     new MkContravariant[F] {
       def safeContramap[A, B](fa: F[A])(f: B => Eval[A]): Eval[F[B]] =
         mkSafeMap(F.fo)(F.unpack(fa))(
           (a: F.I[A]) => mkContraSafe(F.fi)(a)(f)
-        ).map(F.pack)
-    }
-
-  implicit def mkContravariantThriceNested[F[_], I[_], J[_], K[_]](
-    implicit
-    F: Split1.Aux[F, ContraOrMk, ContraOrMk, I, J],
-    G: Split1[J, ContraOrMk, ContraOrMk]
-  ): MkContravariant[F] =
-    new MkContravariant[F] {
-      override def safeContramap[A, B](fa: F[A])(f: B => Eval[A]): Eval[F[B]] =
-        mkContraSafe(F.fo)(F.unpack(fa))(
-          (a: F.I[B]) =>
-            mkContraSafe(G.fo)(G.unpack(a))(
-              (b: G.I[A]) => mkContraSafe(G.fi)(b)(f)
-            ).map(G.pack)
         ).map(F.pack)
     }
 }
@@ -71,13 +79,6 @@ private[derived] abstract class MkContravariantCons extends MkContravariantGener
 }
 
 private[derived] abstract class MkContravariantGeneric {
-  protected type FunctorOrMk[F[_]] = Functor[F] OrElse MkFunctor[F]
-  protected def mkSafeMap[F[_], A, B](F: FunctorOrMk[F])(fa: F[A])(f: A => Eval[B]): Eval[F[B]] =
-    F.unify match {
-      case mk: MkFunctor[F] => mk.safeMap(fa)(f)
-      case p => Eval.later(p.map(fa)(f(_).value))
-    }
-
   protected type ContraOrMk[F[_]] = Contravariant[F] OrElse MkContravariant[F]
 
   protected def mkContraSafe[F[_], A, B](F: ContraOrMk[F])(fa: F[B])(f: A => Eval[B]): Eval[F[A]] =
--- a/core/src/main/scala/cats/derived/contravariant.scala
+++ b/core/src/main/scala/cats/derived/contravariant.scala
@@ -41,7 +41,7 @@ private[derived] abstract class MkContravariantDerivation extends MkContravarian
 }
 
 private[derived] abstract class MkContravariantNested1 extends MkContravariantCons {
-  implicit def mkContravariantFunctorNested[F[_]](implicit F: Split1[F, Functor, ContraOrMk]): MkContravariant[F] =
+  implicit def mkFunctorContravariantNested[F[_]](implicit F: Split1[F, Functor, ContraOrMk]): MkContravariant[F] =
     new MkContravariant[F] {
       def safeContramap[A, B](fa: F[A])(f: B => Eval[A]): Eval[F[B]] =
         Eval.later(F.fo.map(F.unpack(fa))(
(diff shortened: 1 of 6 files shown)


core/src/main/scala/cats/derived/contravariant.scala
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+  implicit def mkContravariantFunctorNested[F[_]](implicit F: Split1[F, FunctorOrMk, ContraOrMk]): MkContravariant[F] =
+    new MkContravariant[F] {
+      def safeContramap[A, B](fa: F[A])(f: B => Eval[A]): Eval[F[B]] =
+        mkSafeMap(F.fo)(F.unpack(fa))(
+          (a: F.I[A]) => mkContraSafe(F.fi)(a)(f)
+        ).map(F.pack)
+    }
+}
```
stephen-lazaro: Seems like `F[G[H[_]]]` should also be contra when `F`, `G`, and `H` are but my attempts to do something like that blew up a bit.

core/src/main/scala/cats/derived/contravariant.scala
```diff
@@ -0,0 +1,95 @@
...
+
+private[derived] abstract class MkContravariantNested extends MkContravariantCons {
+  protected type FunctorOrMk[F[_]] = Functor[F] OrElse MkFunctor[F]
+  protected def mkSafeMap[F[_], A, B](F: FunctorOrMk[F])(fa: F[A])(f: A => Eval[B]): Eval[F[B]] =
+    F.unify match {
+      case mk: MkFunctor[F] => mk.safeMap(fa)(f)
+      case p => Eval.later(p.map(fa)(f(_).value))
+    }
```
stephen-lazaro: This is duplication, do we have a better way to reuse stuff across derivations?

joroKr21:
Hi @stephen-lazaro, welcome to `kittens` and thank you for the PR!

core/src/main/scala/cats/derived/contravariant.scala
joroKr21: Yes, unfortunately you will see some code duplication, we haven't found a good way to get rid of it.

But in this case I would say we are trying to do too much. Kittens already tries to derive as much as it can but only for a single typeclass. E.g. when the user writes `semi.contravariant[F]` we should not try to derive a `Functor`, even just as an intermediate derivation.

Anyway forgive my rambling, what I'm trying to say is that `MkContravariant` shouldn't depend on `MkFunctor`. Instead just adding a `Functor` implicit where necessary is enough.

core/src/main/scala/cats/derived/contravariant.scala
joroKr21: I think here we would hit the limitations of shapeless' `Split1`. I think it would be enough to have two instances for `Contravariant`:
  * `F[_]: Functor, G[_]: ContravariantOrMk => Contravariant[F[G[_]]]`
  * `F[_]: ContravariantOrMk, G[_]: Functor => Contravariant[F[G[_]]]`

And then we might also want to add another one to `MkFunctor` (I think we are missing that):
  * `F[_]: Contravariant, G[_]: Contravariant => Functor[F[G[_]]]`

core/src/main/scala/cats/derived/contravariant.scala
```diff
@@ -0,0 +1,95 @@
...
+private[derived] abstract class MkContravariantGeneric {
+  protected type ContraOrMk[F[_]] = Contravariant[F] OrElse MkContravariant[F]
+
+  protected def mkContraSafe[F[_], A, B](F: ContraOrMk[F])(fa: F[B])(f: A => Eval[B]): Eval[F[A]] =
+    F.unify match {
+      case mk: MkContravariant[F] => mk.safeContramap(fa)(f)
+      case p => Eval.later(p.contramap(fa)(f(_).value))
+    }
+
+  implicit def mkFunctorGeneric[F[_]](implicit F: Generic1[F, MkContravariant]): MkContravariant[F] =
```
joroKr21:
```suggestion
  implicit def mkContraGeneric[F[_]](implicit F: Generic1[F, MkContravariant]): MkContravariant[F] =
```

core/src/test/scala/cats/derived/contravariant.scala
```diff
@@ -0,0 +1,102 @@
...
+  case class Pred[A](run: A => Boolean)
+  case class TwiceNest[A](run: Pred[Pred[A]])
+  case class ThriceNest[A](run: Pred[TwiceNest[A]])
+
+  implicit def eqIntPred[A: Numeric: Choose, B: Eq]: Eq[A => B] = new Eq[A => B] {
+    def diagonal[A](a: A): (A, A) = (a, a)
+    override def eqv(x: A => B, y: A => B): Boolean =
+      Gen.posNum[A]
+        .sample
+        .exists(diagonal[A] _ andThen (x *** y) andThen (Eq[B].eqv _).tupled)
```
joroKr21: It's better to use `MiniInt` and `Boolean` instead of `Int` and `Long` for `ContravariantTests` instead of defining this awkward `Eq` instances for functions. `cats-laws` had equality for functions based on `Arbitrary` but it was removed and I think we shouldn't introduce it if we can avoid it.

core/src/main/scala/cats/derived/contravariant.scala
joroKr21: >   `F[_]: Functor, G[_]: ContravariantOrMk => Contravariant[F[G[_]]]`
>   `F[_]: ContravariantOrMk, G[_]: Functor => Contravariant[F[G[_]]]`

Note that those would have to be prioritized but it doesn't matter in which order.

joroKr21: >  `F[_]: Contravariant, G[_]: Contravariant => Functor[F[G[_]]]`

in `Functor` you could add this in this PR or could be left for later, your choice.

stephen-lazaro: Aha, that makes good sense. That makes a lot more sense than what I tried to do with multiple `Split1` evidences.

core/src/main/scala/cats/derived/contravariant.scala
stephen-lazaro: Nice, ok that makes sense. I'll switch to that.

core/src/test/scala/cats/derived/contravariant.scala
stephen-lazaro: Gotcha, nice, yeah I imagined there was something better than this.

stephen-lazaro:
@joroKr21 I think that incorporates the feedback! Thanks for reading. Let me know if I missed anything or can improve things.

core/src/main/scala/cats/derived/contravariant.scala
joroKr21: Still need to apply this renaming

core/src/main/scala/cats/derived/functor.scala
```diff
@@ -52,6 +52,15 @@ private[derived] abstract class MkFunctorNested extends MkFunctorCons {
       def safeMap[A, B](fa: F[A])(f: A => Eval[B]) =
         mkSafeMap(F.fo)(F.unpack(fa))(mkSafeMap(F.fi)(_)(f)).map(F.pack)
     }
+
+  implicit def mkFunctorFromContraNested[F[_]](implicit F: Split1[F, Contravariant, Contravariant]): MkFunctor[F] =
```
joroKr21: Would mind adding a test for this? Something like:
```scala
type Pred[-A] = A => Boolean
Functor[Lambda[a => Pred[Pred[a]]]]
```

I'm just worried about implicit resolution. It looks like it should be ambiguous with the `mkFunctorNested` method :thinking:

core/src/main/scala/cats/derived/contravariant.scala
```diff
@@ -0,0 +1,88 @@
...
+  implicit val mkContraCNil: MkContravariant[Const[CNil]#λ] = mkContraConst
+
+  implicit def mkContraConst[T]: MkContravariant[Const[T]#λ] =
+    new MkContravariant[Const[T]#λ] {
+      def safeContramap[A, B](t: T)(f: B => Eval[A]): Eval[T] = Eval.now(t)
+    }
+}
+
+private[derived] abstract class MkContravariantNested extends MkContravariantCons {
+  implicit def mkContravariantFunctorNested[F[_]](implicit F: Split1[F, Functor, ContraOrMk]): MkContravariant[F] =
```
joroKr21:
```suggestion
  implicit def mkFunctorContravariantNested[F[_]](implicit F: Split1[F, Functor, ContraOrMk]): MkContravariant[F] =
```
I think we are still missing (hence the suggestion):
```scala
implicit def mkContravariantFunctorNested[F[_]](implicit F: Split1[F, ContraOrMk, Functor]): MkContravariant[F]
```

joroKr21:
@stephen-lazaro almost there :sweat_smile:

core/src/main/scala/cats/derived/contravariant.scala
stephen-lazaro: Huh, so weird, I viscerally remember doing this, must have git fu-d myself into a snafu, thanks.

core/src/main/scala/cats/derived/functor.scala
stephen-lazaro: You know, I've been playing with that and I'm not sure it's quite right, actually. I'll mess with the tests a bit and make sure that it works correctly.

stephen-lazaro:
Cool! I think that might be about right @joroKr21?

core/src/main/scala/cats/derived/contravariant.scala
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+private[derived] abstract class MkContravariantNested0 extends MkContravariantGeneric {
+  implicit def mkContravariantFunctorNested[F[_]](implicit F: Split1[F, ContraOrMk, Functor]): MkContravariant[F] =
+    new MkContravariant[F] {
+      def safeContramap[A, B](fa: F[A])(f: B => Eval[A]): Eval[F[B]] =
+        mkContraSafe(F.fo)(F.unpack(fa))(
+          (a: F.I[B]) => Eval.later(F.fi.map(a)(f andThen (_.value)))
+        ).map(F.pack[B])
+    }
```
stephen-lazaro: Interestingly, if this instance is higher in the chain, it sends the compiler into an infinite loop during autoresolution tests, which is kinda interesting.

stephen-lazaro:
Ah bummer, that only works on 2.13. I'll see if I can fix it.

core/src/main/scala/cats/derived/contravariant.scala
joroKr21: Hmm, does it have anything to do with Id?

stephen-lazaro: Possibly? I end up seeing a loop in the implicit log around various Generic10, Generic11, and Generic13 instances IIRC? I'm gonna mess around with 2.12/2.11 a bit more and see if I can get it to work on there.

joroKr21: Sorry that I haven't found time to help you, I'm quite busy these two weeks. Next week I will probably be able to.

stephen-lazaro: No problem! I didn't really expect it to be honest, I just like to keep these things updated :)

joroKr21:
@stephen-lazaro I'm sorry you encountered this weird bug (is it even a bug?)

I tried to but didn't manage to isolate the source of the seemingly endless recursion in implicits resolution. Wdyt about just dropping the nested instances for now and we can merge this?

joroKr21:
I squashed the commits from this PR and removed the broken instance in #179

joroKr21:
@stephen-lazaro I incorporated your commits into #179 and did a few tweaks on top as well as removed that problematic instance. Thank you so much and sorry for the delay!

stephen-lazaro:
@joroKr21 Awesome! No worries. Thanks for carrying it over the line!

PR #172, opened 2019-07-19
Fix libraryDependencies for scala.js

Taig:
The dependencies are not properly referenced for scalajs causing linking errors

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import com.typesafe.sbt.pgp.PgpKeys.publishSigned
 import org.scalajs.sbtplugin.cross.CrossProject
 import ReleaseTransformations._
 import sbt._
-import sbtcrossproject.{CrossType, crossProject}
+import sbtcrossproject.CrossPlugin.autoImport.{CrossType, crossProject}
 
 lazy val buildSettings = Seq(
   organization := "org.typelevel",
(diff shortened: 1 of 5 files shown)


joroKr21:
Huh, I wonder how it was working before :thinking: 

You need to update `EmptySuite`, looks like `scala.js` throws a different exception type.

Taig:
Ah, I thought that was a flaky or something. I'll update the PR soon.

Taig:
Tests work fine on my machine™. I'll try to find out what's going wrong here.

joroKr21:
As far as I could see on Travis the test that expects a StackOverflow doesn't pass on scala.js because the exception type is different. I don't know anything about exceptions in scala.js

Taig:
I tried to avoid platform specific code. Let me know if you'd rather have another approach.

joroKr21:
Thanks, it would probably just be enough to expect any exception, this is quite the corner case, but it's also LGTM as it is.

joroKr21:
I'm hesitant to merge this because the build now takes forever. Also running `test` locally is too slow. I guess now we have the problem of too many tests :smile: 

@kailuowang, @LukaJCB WDYT?

kailuowang:
So just the confirm my understanding: the build didn't catch the scalajs link error and didn't run scalajs tests? The build takes a lot longer now because it's running tests on scalajs?

joroKr21:
Yes I think that's exactly the case. Very odd.

kailuowang:
I hope we can fix the build so that scalajs link error in the future would be captured, in case we introduce another dependency and use the wrong setting (`%%` instead of `%%%`), but I guess that we can do that in another PR. 
For this one I am not sure if we should merge as is either. 
We should try use a different settings on scalajs settings. 
One quick thing to try is to use the [TestSettings in Cats](https://github.com/typelevel/cats/blob/master/testkit/src/main/scala/cats/tests/CatsSuite.scala#L16-L28), it's readly available through `cats-testkit`

joroKr21:
@Taig I hope you don't mind, I squashed your commits and added one commit to use `TestSettings` from Cats which reduce the build time significantly.

joroKr21:
Looks good now, thanks @Taig!

Taig:
👍 Do you have any further reading on why that change reduces build times so significantly?

joroKr21:
JavaScript is slow :smile: take a look at the settings, everything is dialled down for Scala.js:
```scala
    PropertyCheckConfiguration(
      minSuccessful = if (Platform.isJvm) PosInt(50) else PosInt(5),
      maxDiscardedFactor = if (Platform.isJvm) PosZDouble(5.0) else PosZDouble(50.0),
      minSize = PosZInt(0),
      sizeRange = if (Platform.isJvm) PosZInt(10) else PosZInt(5),
      workers = if (Platform.isJvm) PosInt(2) else PosInt(1)
    )
```

PR #176, opened 2019-08-29
Add Gen.delay to Arbitrary[Recursive]

ashawley:
ScalaCheck will simplify its implementation of `Arbitrary[Option]` in the next release 1.14.1, but the change incidentally increases the probability of a stackoverflow for recursive data types with `Option`.

I've tested this with a local snapshot of ScalaCheck.

https://github.com/typelevel/scalacheck/issues/488

### DIFF
--- a/core/src/test/scala/cats/derived/adtdefns.scala
+++ b/core/src/test/scala/cats/derived/adtdefns.scala
@@ -56,7 +56,7 @@ object TestDefns {
   object Recursive extends ((Int, Option[Recursive]) => Recursive) {
 
     implicit val arbitrary: Arbitrary[Recursive] =
-      Arbitrary(Arbitrary.arbitrary[(Int, Option[Recursive])].map(tupled))
+      Arbitrary(Gen.delay(Arbitrary.arbitrary[(Int, Option[Recursive])]).map(tupled))
 
     implicit val cogen: Cogen[Recursive] =
       Cogen[(Int, Option[Recursive])].contramap(unapply(_).get)
--- a/core/src/test/scala/cats/derived/adtdefns.scala
+++ b/core/src/test/scala/cats/derived/adtdefns.scala
@@ -56,7 +56,7 @@ object TestDefns {
   object Recursive extends ((Int, Option[Recursive]) => Recursive) {
 
     implicit val arbitrary: Arbitrary[Recursive] =
-      Arbitrary(Gen.delay(Arbitrary.arbitrary[(Int, Option[Recursive])]).map(tupled))
+      Arbitrary(Gen.delay(Arbitrary.arbitrary[(Int, Option[Recursive])].map(tupled)))
 
     implicit val cogen: Cogen[Recursive] =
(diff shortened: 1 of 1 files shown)


kailuowang:
travis is being funky. We can change to openjdk in .travis to soothe it.

ashawley:
Ok, switching to openjdk is in #177

joroKr21:
@ashawley thanks, I completely forgot about this :+1: 

Hmm, now that `Gen` doesn't stack overflow, tests of stack unsafe methods do.
Maybe just remove those tests?

ashawley:
I didn't see those tests come up while working on this.  Where are those?

joroKr21:
> I didn't see those tests come up while working on this. Where are those

Check travis, there is an `Eq[Recursive]` test.

ashawley:
Ah, ok. I see this failure for 2.13 in Travis:

```
[info] - auto.Eq[Recursive].eq.antisymmetry *** FAILED ***
[info]   NullPointerException was thrown during property evaluation.
[info]     Message: "None"
[info]     Occurred when passed generated values (
[info]       arg0 = Recursive(-1,Some(Recursive(2147483647,None))),
[info]       arg1 = Recursive(-1,Some(Recursive(2147483647,None))),
[info]       arg2 = org.scalacheck.GenArities$$Lambda$6893/583834148@24a41151
[info]     )
```

I can't reproduce locally, though.  Which test should I delete?

joroKr21:
Ok let me try to reproduce it locally, although I may not have time to do that today.

joroKr21:
I had to add `Gen.delay` in #178 (not sure why), let's see how it goes.

joroKr21:
@ashawley maybe the difference is adding `delay` after we `map` not before?

ashawley:
And #178 is for unrelated changes from ScalaCheck?  So this is already a temperamental test case?

joroKr21:
> So this is already a temperamental test case?

Nope, the missing `Gen.delay` was always a problem, but depending on the stack size and chance it hasn't occurred until now.

joroKr21:
> Nope, the missing Gen.delay was always a problem, but depending on the stack size and chance it hasn't occurred until now.

Ok, scratch that, master is still fails sometimes. The problem is - even if we build a `Recursive` instance in a stack-safe way, methods that work with it and are not stack-safe will blow up.

ashawley:
Weird, so it is a temperamental test case all of a sudden.

We actually realized that the forthcoming implementation of `Arbitrary[Option[T]]` in ScalaCheck will reveal problems like this, but we also recognized that the prior implementation had the same defect of stackoverflow, there was just a lower risk of it happening.

So I'm not sure you can take much stock in where you put the `Gen.delay` if this is a flakey test.  In theory, the way `Arbitrary.apply` accepts an expression is supposed to be call-by-name.

ashawley:
> Ok, scratch that, master is still fails sometimes.

Yeah, this is a deeper problem than it looks.

joroKr21:
I believe the correct solution is to use sized generators (as if you were generating a binary tree for example): see #180

PR #177, opened 2019-08-29
Switch to openjdk8 in Travis

ashawley:


ashawley:
The build is green again

PR #180, opened 2019-08-31
Make Gen[Recursive] sized for stack safety

joroKr21:
Similar to `Gen[Tree[A]]` in the same file.

### DIFF
--- a/core/src/test/scala/cats/derived/adtdefns.scala
+++ b/core/src/test/scala/cats/derived/adtdefns.scala
@@ -55,8 +55,13 @@ object TestDefns {
   final case class Recursive(i: Int, is: Option[Recursive])
   object Recursive extends ((Int, Option[Recursive]) => Recursive) {
 
-    implicit val arbitrary: Arbitrary[Recursive] =
-      Arbitrary(Gen.delay(Arbitrary.arbitrary[(Int, Option[Recursive])].map(tupled)))
+    implicit val arbitrary: Arbitrary[Recursive] = {
+      def recursive(size: Int): Gen[Recursive] = for {
+        i <- Arbitrary.arbitrary[Int]
+        is <- if (size <= 0) Gen.const(None) else Gen.option(recursive(size / 2))
+      } yield Recursive(i, is)
+      Arbitrary(Gen.sized(recursive))
+    }
 
     implicit val cogen: Cogen[Recursive] =
       Cogen[(Int, Option[Recursive])].contramap(unapply(_).get)
@@ -129,7 +134,7 @@ object TestDefns {
       val leaf = Arbitrary.arbitrary[A].map(Leaf.apply)
 
       def tree(maxDepth: Int): Gen[Tree[A]] =
-        if (maxDepth == 0) leaf
+        if (maxDepth <= 0) leaf
         else Gen.oneOf(leaf, node(maxDepth))
 
       def node(maxDepth: Int): Gen[Tree[A]] = for {


joroKr21:
This should fix the flakyness on master

ashawley:
Yeah, this is a better approach. 👍

core/src/test/scala/cats/derived/adtdefns.scala
```diff
@@ -55,8 +55,19 @@ object TestDefns {
...
+    implicit val arbitrary: Arbitrary[Recursive] = {
+      def tail(maxLength: Int): Gen[Option[Recursive]] =
+        if (maxLength <= 0) Gen.const(None)
+        else Gen.option(recursive(maxLength))
+
+      def recursive(maxLength: Int): Gen[Recursive] = for {
+        i <- Arbitrary.arbitrary[Int]
+        n <- Gen.choose(0, maxLength - 1)
+        is <- tail(n)
+      } yield Recursive(i, is)
```
ashawley: There's a good chance you only need one method for recursion.

joroKr21:
Wow, that's a first for me:
`[error] java.util.concurrent.ExecutionException: java.lang.InternalError: Malformed class name`

core/src/test/scala/cats/derived/adtdefns.scala
joroKr21: right :+1:

core/src/test/scala/cats/derived/adtdefns.scala
```diff
@@ -55,8 +55,14 @@ object TestDefns {
   final case class Recursive(i: Int, is: Option[Recursive])
   object Recursive extends ((Int, Option[Recursive]) => Recursive) {
 
-    implicit val arbitrary: Arbitrary[Recursive] =
-      Arbitrary(Gen.delay(Arbitrary.arbitrary[(Int, Option[Recursive])].map(tupled)))
+    implicit val arbitrary: Arbitrary[Recursive] = {
+      def recursive(maxLength: Int): Gen[Recursive] = for {
+        i <- Arbitrary.arbitrary[Int]
+        n <- Gen.choose(0, maxLength - 1)
+        is <- if (n <= 0) Gen.const(None) else Gen.option(recursive(n))
```
ashawley: Yes, this works, but I'd probably just do `recursive(n / 2)` rather than `Gen.choose`. It depends how much nesting you want to test versus how much you want to avoid stackoverflow.

Here's an alternate implementation.

```scala
      def recur(depth: Int): Gen[Recursive] =
        Arbitrary.arbitrary[Int].flatMap { i =>
          if (depth > 0) {
            recur(depth / 2).map(r => Recursive(i, Some(r)))
          } else {
            Gen.const(Recursive(i, None))
          }
        }
      Arbitrary(Gen.sized(sz => recur(sz)))
```

core/src/test/scala/cats/derived/adtdefns.scala
```diff
@@ -55,8 +55,14 @@ object TestDefns {
...
+      } yield Recursive(i, is)
+      Arbitrary(recursive(32))
```
ashawley: Oh, I see.  You've dropped `Gen.sized` and just hard-coded the depth.

core/src/test/scala/cats/derived/adtdefns.scala
joroKr21: Yeah I was thinking whether not to halve the size instead. It should be more robust :+1:

core/src/test/scala/cats/derived/adtdefns.scala
joroKr21: Yes but I like your suggestion more.

core/src/test/scala/cats/derived/adtdefns.scala
joroKr21: But I still want to use `Gen.option` for more randomization.

ashawley: The default size for `Gen.sized` is randomly chosen between 0 and 99.  Yes, you would be getting more randomization, but we're halving with `n / 2` so there are only 8 max depths of `Recursive`.    So maybe `Gen.option` would just distribute the occurrence of the 8 depths more evenly?  It would take experimentation.

ashawley:
Thanks for take care of this, Georgi.  This should fix the Scala Community Build and also make the upgrade here to ScalaCheck 1.14.1 easier.

joroKr21:
Ok this test is still flaky: https://travis-ci.org/typelevel/kittens/jobs/580757020
I'm out of ideas at this point.
Should we just give up on testing that particular class?

ashawley:
It's an NPE, not a stackoverflow.  So something else seems awry.  What is the test `auto.Order[Recursive].order.antisymmetry` about?  It seems to be consistently failing, but the other tests that also use `Recursive` don't.

As i said previously, I don't think the depth should be any more than 8, so it seems like the stack safety concern is resolved.  I don't know enough about kittens to really grok what else could be failing here.

joroKr21:
@ashawley I think this time it's because of the `Cogen` instance. See #187

PR #182, opened 2019-09-03
Derive Invariant

stephen-lazaro:
Resolves https://github.com/typelevel/kittens/issues/144

### DIFF
--- a/core/src/main/scala/cats/derived/invariant.scala
+++ b/core/src/main/scala/cats/derived/invariant.scala
@@ -26,7 +26,7 @@ import scala.annotation.implicitNotFound
 trait MkInvariant[F[_]] extends Invariant[F] {
   def safeImap[A, B](fa: F[A])(g: A => Eval[B])(f: B => Eval[A]): Eval[F[B]]
   def imap[A, B](fa: F[A])(g: A => B)(f: B => A): F[B] =
-    safeImap(fa)(g andThen Eval.now)(f andThen Eval.now).value
+    safeImap(fa)(a => Eval.later(g(a)))(b => Eval.later(f(b))).value
 }
 object MkInvariant extends MkInvariantDerivation {
   def apply[F[_]](implicit F: MkInvariant[F]): MkInvariant[F] = F
--- a/core/src/main/scala/cats/derived/invariant.scala
+++ b/core/src/main/scala/cats/derived/invariant.scala
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2015 Miles Sabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats
+package derived
+
+import shapeless._
+import util.VersionSpecific.OrElse
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not derive an instance of Invariant[${F}]")
+trait MkInvariant[F[_]] extends Invariant[F] {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/derived/invariant.scala
```diff
@@ -0,0 +1,91 @@
...
+import shapeless._
+import util.VersionSpecific.OrElse
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not derive an instance of Invariant[${F}]")
+trait MkInvariant[F[_]] extends Invariant[F] {
+  def safeImap[A, B](fa: F[A])(g: A => Eval[B])(f: B => Eval[A]): Eval[F[B]]
+  def imap[A, B](fa: F[A])(g: A => B)(f: B => A): F[B] =
+    safeImap(fa)(g andThen Eval.now)(f andThen Eval.now).value
```
joroKr21: I wonder what's the difference between using `f andThen Eval.now` and `a => Eval.later(f(a))`.
But I would say let's use the latter since it would be the same as `MkFunctor`.

core/src/main/scala/cats/derived/invariant.scala
```diff
@@ -0,0 +1,91 @@
...
+
+private[derived] abstract class MkInvariantNested extends MkInvariantCons {
+  implicit def mkFunctorInvariantNested[F[_]](implicit F: Split1[F, Invariant, InvariantOrMk]): MkInvariant[F] =
+    new MkInvariant[F] {
+      def safeImap[A, B](fa: F[A])(g: A => Eval[B])(f: B => Eval[A]): Eval[F[B]] =
+        Eval.later(F.fo.imap(F.unpack(fa))(
+          (a: F.I[A]) => mkImapSafe(F.fi)(a)(g)(f).value
+        )(
+          (b: F.I[B]) => mkImapSafe(F.fi)(b)(f)(g).value
+        )).map(F.pack[B])
```
joroKr21: This can be implemented without calling `.value` (it's a bit tricky):
```scala
      def safeImap[A, B](fa: F[A])(f: A => Eval[B])(g: B => Eval[A]) =
        mkSafeImap(F.fo)(F.unpack(fa))(mkSafeImap(F.fi)(_)(f)(g))(mkSafeImap(F.fi)(_)(g)(f)).map(F.pack)
```

joroKr21:
You might need to rebase manually, I don't know why github is saying there are conflicts, I don't see any.

core/src/main/scala/cats/derived/invariant.scala
stephen-lazaro: Ah good point, I think we would need the latter. Thanks.

core/src/main/scala/cats/derived/invariant.scala
stephen-lazaro: oh that's pretty slick actually, i'll switch to that.

stephen-lazaro:
Great, thanks @joroKr21. I believe I've now incorporated the feedback.

stephen-lazaro:
Odd... looks like an unrelated test failed for `Recursive` in `2.11`. I'll investigate.

stephen-lazaro:
Yep, looks like a flappy test.

core/src/test/scala/cats/derived/adtdefns.scala
```diff
@@ -77,6 +77,18 @@ object TestDefns {
...
+  object Bivariant {
+      implicit def arbitrary[A: Arbitrary]: Arbitrary[Bivariant[A]] = Arbitrary(
+        for {
+          a <- Arbitrary.arbitrary[A]
+          f <- Arbitrary.arbitrary[Boolean].map(Function.const[Boolean, A])
+        } yield Bivariant[A](f, a)
+      )
+  }
+
+  case class Pred[A](run: A => Boolean)
```
joroKr21: Is this used anywhere?

core/src/test/scala/cats/derived/adtdefns.scala
```diff
@@ -77,6 +77,18 @@ object TestDefns {
       Cogen[(Int, T, Long, List[T], String)].contramap(unapply(_).get)
   }
 
+  case class Bivariant[A](run: A => Boolean, store: A)
```
joroKr21: Hmm, this is not really bivariant. Bivariance (also called sometimes "phantom variance") is an interesting concept: https://typelevel.org/blog/2016/09/19/variance-phantom.html. It can be both covariant and contravariant at the same time. Anyway just wanted to mention that, it's just a test class.

joroKr21:
> Yep, looks like a flappy test.

This should fix it: #187

PR #189, opened 2019-09-08
Improve implicitNotFound messages

### LINKED ISSUE #89: Improve @implicitNotFound error messages
I copied the code from documentation and replaced Show with Eq:
```
import cats._

case class A(a: Int)

object Test {
  implicit val eq: Eq[A] = {
    import derived.auto.eq._
    derived.semi.eq
  }
}
```
Compiled with Scala 2.12.4 (https://scastie.scala-lang.org/z4Ksxl4lR26ECCDjmGSAxQ)
```
Adaptation of argument list by inserting () is deprecated: this is unlikely to be what you want.
        signature: OrElse.primary[A, B](implicit a: A): shapeless.OrElse[A,B]
  given arguments: 
 after adaptation: OrElse.primary((): Unit)

ambiguous implicit values:
 both value mkEqHnil in class MkEqDerivation of type => cats.derived.MkEq[shapeless.HNil]
 and value mkEqCnil in class MkEqDerivation of type => cats.derived.MkEq[shapeless.CNil]
 match expected type cats.derived.MkEq[A]

could not find implicit value for parameter ev: cats.derived.MkEq[A]

not enough arguments for method eq: (implicit ev: cats.derived.MkEq[A])cats.Eq[A].
Unspecified value parameter ev.
```

joroKr21:
Putting this up for discussion. How exactly should we format and formulate the messages?

Current proposal:
  * One sentence to mention what exactly was not derived
  * One sentence to introduce a list of conditions
  * List the conditions, one of which must be satisfied

~Also: should we use kind-projector syntax or not?~

Closes #89

### DIFF
--- a/core/src/main/scala/cats/derived/applicative.scala
+++ b/core/src/main/scala/cats/derived/applicative.scala
@@ -22,7 +22,15 @@ import util.VersionSpecific.OrElse
 
 import scala.annotation.implicitNotFound
 
-@implicitNotFound("Could not derive an instance of Applicative[${F}]")
+@implicitNotFound("""
+Could not derive an instance of Applicative[F] where F = ${F}.
+Make sure that F[_] satisfies one of the following conditions:
+  * it is a constant type λ[x => T] where T: Monoid
+  * it is a nested type λ[x => G[H[x]]] where G: Applicative and H: Applicative
+  * it is a generic case class where all fields have an Applicative instance
+
+Note: using kind-projector notation - https://github.com/typelevel/kind-projector
+""".trim)
 trait MkApplicative[F[_]] extends Applicative[F]
 
 object MkApplicative extends MkApplicativeDerivation {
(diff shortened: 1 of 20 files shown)


kailuowang:
@joroKr21 that proposed message looks very help to me. 
I don't have a strong opinion on whether we should use kind-projector syntax. I am thinking maybe using the syntax but with a quick note linking to kind-projector would be helpful.

joroKr21:
@kailuowang what did you mean by that:
> that proposed message looks very help to me

kailuowang:
sorry @joroKr21, I meant "very helpful". was on my phone.

joroKr21:
@kailuowang how does it look now?

kailuowang:
Lgtm!

joroKr21:
@kailuowang updated all typeclass derivations. Don't know if we can think of better messages for things that are not typeclasses.

Issue #142, opened 2019-04-27
NonEmptyTraverse and Reducible derivation


joroKr21:
If `Traverse` and `Foldable` are possible, so should `NonEmptyTraverse` and `Reducible`.
But at this point code reuse between derivations might be necessary.

PR #211, opened 2019-12-05
Reducible derivation

joroKr21:
Due to the interaction between `MkFoldable` and `MkReducible`,
`MkReducible` had to extend `MkFoldable` to be stack safe.
An alternative would be to add a `MkReducible` case in
`mkSafeFoldLeft`.

Part of #142


joroKr21:
I'm not sure why the tests pass locally but fail on Travis :thinking:

PR #212, opened 2019-12-07
Deprecate auto in favor of semiauto

joroKr21:
`semiauto` doesn't use `Lazy`.

That improves compile time performance marginally and has better error messages on Scala 2.13.
Moved `semiauto` test instances to companion objects to ensure serializability.

This would have to go into a 2.1.0 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ cat: Cat[Int] = Cat(1,List(2, 3))
 ```scala
 scala> implicit val fc: Functor[Cat] = { 
           import auto.functor._           
-          semi.functor }
+          semiauto.functor }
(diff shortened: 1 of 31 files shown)


joroKr21:
Ahh, looks like unlike by-name implicits, shapeless' `Lazy` doesn't work in this case :disappointed: 

So we're stuck with the degraded error messages on 2.13 I'm afraid.

joroKr21:
New idea - we can make `semiauto` version specific. On 2.11 and 2.12 it will be an alias to `semi` and on 2.13 it will be without `Lazy` - not sure if this is a good idea or not.

kailuowang:
That sounds good to me.

PR #215, opened 2019-12-11
Add parSequence for hlists

### LINKED ISSUE #290: Traverser doesn't handle HNil
The following examples shows that `Traverser` doesn't handle `HNil` properly while `Mapper`/`Sequencer` do well. 
Is this a design error of Traverser not to include a traversable `F[_]` type parameter like `Sequencer` does?

```scala
import cats.implicits._
import cats.sequence._
import shapeless._
import shapeless.ops.hlist.Mapper
object toOptionPoly extends Poly1 {
  implicit def kase[T]: Case.Aux[T, Option[T]] = at(_.some)
}
def traverse[T <: HList, TraverserOut](v: T)(
    implicit traverser: Traverser.Aux[T, toOptionPoly.type, TraverserOut]
  ): TraverserOut = traverser(v)
def mapAndSequence[T <: HList, MapperOut <: HList, SequencerOut <: HList](v: T)(
    implicit mapper: Mapper.Aux[toOptionPoly.type, T, MapperOut],
    sequencer: Sequencer.Aux[MapperOut, Option, SequencerOut]
  ): Option[SequencerOut] = sequencer(mapper(v))

traverse(1 :: HNil)
// implicit not found
traverseHList(HNil: HNil)

mapAndSequence(1 :: HNil)
// works fine
mapAndSequence(HNil: HNil)
```

travisbrown:
As discussed [here](https://stackoverflow.com/a/59266384/334519), this allows things like error accumulation when validating with `Either`.

This could be done without changing `Sequence`, just as syntax (via the approach in the Stack Overflow answer), but that involves a lot of extra overhead (of the syntactic, runtime, and compile-time variety). 

I would have added a record version and `parTraverse`, but both `RecordSequence` and `Traverse` don't preserve the `F[_]` context, which makes it a little more complicated (I can understand the reasoning for `Traverse` not having an `F[_]`, but not `RecordSequence`, which seems like it should match `Sequence`'s type members).

Also fixes #290


joroKr21:
That was way shorter than I expected. I still need to reacquaint myself with this part of `kittens` but what's with the binary incompatibility? Looks like `protected[sequence]` doesn't help?

travisbrown:
@joroKr21 Adding an abstract method to a trait always breaks bincompat, so `par` is a problem on all Scala versions, and even `parApply` is a problem on 2.11. This could be done in a bincompat way using the approach in the Stack Overflow answer, but I think it'd be better to park it until the next breaking version update.

joroKr21:
> I think it'd be better to park it until the next breaking version update.

Sounds good to me. I was going to suggest the `Aux` change but I see you already did it anyway.
If we will be waiting for next breaking release we can also figure out `Traverse`. But we have time.
We will probably have to do a breaking release after `shapeless` which will be coming soon I hope.

sorenbug:
[I wrote `ParTraverse` as a typeclass](https://mystb.in/afogukicot.scala) without knowing about this PR until now. Should I do anything with it (i.e. open a separate PR for it) or is it going to be implemented like how it is for `Sequence` (at the call site)?

joroKr21:
I think it's reasonable to implement it the same way as `parSequence` but due to binary incompatibility would have to wait for the next major version.

joroKr21:
I added `parTraverse` as well. I can push to your branch if you enable the setting or open a PR to it?

travisbrown:
@joroKr21 You should be able to push now?

joroKr21:
Didn't manage to push 🤷‍♂ I created a PR to your branch instead.

joroKr21 (approved):

joroKr21:
Apologies for not realising sooner - we dropped 2.11, the traits are sealed and we provide a default sequential implementation and override it in implementations - therefore binary compatibility is not a problem.

joroKr21:
On no, Scala 2.12 😞

PR #217, opened 2019-12-15
NonEmptyTraverse derivation

### LINKED ISSUE #142: NonEmptyTraverse and Reducible derivation

joroKr21:
Had to modify `MkTraverse` to extend `MkFoldable`.

Credit goes to @LukaJCB for the idea to use `Either` as the carrier to type to get an `Applicative` when we have only `G[_]: Apply`.

Closes #142

### DIFF
--- a/core/src/main/scala/cats/derived/traverse.scala
+++ b/core/src/main/scala/cats/derived/traverse.scala
@@ -14,16 +14,15 @@ Make sure that F[_] satisfies one of the following conditions:
   * it is a generic sealed trait where all subclasses have a Traverse instance
 
 Note: using kind-projector notation - https://github.com/typelevel/kind-projector""")
-trait MkTraverse[F[_]] extends Traverse[F] {
+trait MkTraverse[F[_]] extends Traverse[F] with MkFoldable[F] {
   def safeTraverse[G[_]: Applicative, A, B](fa: F[A])(f: A => Eval[G[B]]): Eval[G[F[B]]]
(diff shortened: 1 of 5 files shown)


joroKr21:
Argh, still not stack-safe, any ideas? :thinking:

core/src/main/scala/cats/derived/traverse.scala
```diff
@@ -14,16 +14,15 @@ Make sure that F[_] satisfies one of the following conditions:
...
-  def safeFoldLeft[A, B](fa: F[A], b: B)(f: (B, A) => Eval[B]): Eval[B]
   def foldRight[A, B](fa: F[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B]
 
   def traverse[G[_]: Applicative, A, B](fa: F[A])(f: A => G[B]): G[F[B]] =
     safeTraverse(fa)(a => Eval.later(f(a))).value
 
-  def foldLeft[A, B](fa: F[A], b: B)(f: (B, A) => B): B =
-    safeFoldLeft(fa, b)((b, a) => Eval.later(f(b, a))).value
+  override def foldLeft[A, B](fa: F[A], b: B)(f: (B, A) => B) =
+    super[MkFoldable].foldLeft(fa, b)(f)
```
joroKr21: This makes it binary compatible :eyes:

PR #222, opened 2020-02-23
drop Scala 2.11, bump dependencies

larsrh:


joroKr21:
What's the motivation for this?

larsrh:
It follows similar steps for the other libraries in the Cats ecosystem, since Cats dropped 2.11 support a while ago.

joroKr21:
> It follows similar steps for the other libraries in the Cats ecosystem, since Cats dropped 2.11 support a while ago.

But I'm not convinced we should do this right now. Cats has a legitimate reason - maintaining binary compatibility with 2.11 was too burdensome. But here we don't have this problem.

kailuowang:
I am mostly with @joroKr21 on this one. However, will adding support for Scala.js 1.0 be more difficult with 2.11 support?

larsrh:
I don't know, are there any plans for 2.11+1.0 support in Cats? I think this may be too difficult.

joroKr21:
> I don't know, are there any plans for 2.11+1.0 support in Cats? I think this may be too difficult.

Ah, well that would be a legitimate reason 😄 (if Cats will not support SJS 1.0 on 2.11)

larsrh:
Good point, I'll do that.

joroKr21:
@kailuowang do you think we should merge this and publish 2.0 for Scala.js 1.0 or go directly for 2.1

larsrh:
I'd personally recommend back-publishing a sjs 1 artifact, and then drop sjs 0.6 going forward.

PR #224, opened 2020-03-15
Update dependencies and switch to ScalaJS 1.0

joroKr21:


joroKr21:
@kailuowang I think we are ready to release 2.0.1 🙏

PR #234, opened 2020-04-23
Fixes typo artefacts -> artifacts

Steven0351:


milessabin:
Not a typo.

joroKr21:
> Not a typo.

Technically that's true but I think going back to Maven the norm is "artifact":
http://maven.apache.org/ref/3.6.3/maven-artifact/

PR #246, opened 2020-08-22
Sequence InvariantMonoidals

vinayakpathak:
I don't think we need `F[_]` to be an `Applicative` to be able to sequence over it. `InvariantMonoidal` is all we need.


vinayakpathak:
Yes, ok, I wasn't sure how to take care of the binary incompatibility. I'll try what you suggest and update

vinayakpathak:
Ok, I've updated it. Seems like it's binary compatible now. I also added a few tests to test just the `InvariantMonoidal` part.

I'm not 100% sure that the way I've split things into different priorities is right. Please let me know if I've messed something up.

core/src/main/scala/cats/sequence/sequence.scala
```diff
@@ -17,85 +17,157 @@ trait Sequencer[L <: HList] extends Serializable {
...
+}
 
+private[sequence] abstract class MkHConsSequencerLowPrio extends MkHNilSequencerLowPrio {
   type Aux[L <: HList, F0[_], LOut0] = Sequencer[L] {
     type F[X] = F0[X]
     type LOut = LOut0
   }
 
+  implicit def mkHConsSequencerForInvariant[F0[_], H, FT <: HList, T <: HList](
+                                                                                implicit tailSequencer: Aux[FT, F0, T], F: InvariantMonoidal[F0]
```
joroKr21, line 51:
```suggestion
                                                                                implicit tailSequencer: Aux[FT, F0, T], F: InvariantSemigroupal[F0]
```

core/src/main/scala/cats/sequence/sequence.scala
```diff
@@ -17,85 +17,157 @@ trait Sequencer[L <: HList] extends Serializable {
   type F[_]
   type LOut <: HList
   type Out = F[LOut]
+
   def apply(hl: L): Out
 }
 
-private[sequence] abstract class MkHConsSequencer {
+trait MkHNilSequencerLowPrio {
```
joroKr21, line 24: This is to ensure we can evolve it later while preserving binary compatibility:
```suggestion
private[sequence] abstract class MkHNilSequencerLowPrio {
```

core/src/main/scala/cats/sequence/sequence.scala
```diff
@@ -17,85 +17,157 @@ trait Sequencer[L <: HList] extends Serializable {
...
 
 trait GenericSequencer[L <: HList, T] extends Serializable {
   type Out
+
   def apply(hl: L): Out
 }
 
-object GenericSequencer {
-  type Aux[L <: HList, T, Out0] = GenericSequencer[L, T] { type Out = Out0 }
+trait GenericSequencerInstanceLowPrio {
```
joroKr21, line 136: This is to ensure we can evolve it later while preserving binary compatibility:
```suggestion
private[sequence] abstract class GenericSequencerInstanceLowPrio {
```

core/src/main/scala/cats/sequence/sequence.scala
```diff
@@ -17,85 +17,157 @@ trait Sequencer[L <: HList] extends Serializable {
...
 @implicitNotFound("cannot construct sequencer, make sure that every field value of you record ${L} is an Apply")
 trait RecordSequencer[L <: HList] extends Serializable {
   type Out
+
   def apply(record: L): Out
 }
 
-object RecordSequencer {
-  type Aux[L <: HList, Out0] = RecordSequencer[L] { type Out = Out0 }
+trait RecordSequencerInstanceLowPrio {
```
joroKr21, line 100: This is to ensure we can evolve it later while preserving binary compatibility:
```suggestion
private[sequence] abstract class RecordSequencerInstanceLowPrio {
```

joroKr21 (changes requested):
Thanks for the contribution! This is really complex stuff, but the use case with `Semigroup` looks really nice (although we already have derivation for that particular type class). The inheritance relation between `MkHNilSequencerLowPrio` and `MkHConsSequencerLowPrio` also has to be inverted.

joroKr21 (approved):
I made the changes because I'd like to merge and release 2.2.0 this weekend

vinayakpathak:
Oh, thanks for merging this!

My original motivation for this was to be able to write some appropriate typeclass instances for probability distributions. So you can define a `Rand[T]` class to represent a probability distribution and then you can define a monad instance for `Rand`. Combined with kittens, this gives you a great way to define joint distributions. For example, if you have a `Rand[A]` and a `Rand[B]` you can use the type-level sequence to get a `Rand[A :: B :: HNil]` for free, which is awesome!

However, with the previous code, this would only work if you had an `Applicative` instance for `Rand`. When doing probabilistic modelling, you face situations where all you have is a `Density[A]` (i.e., an unnormalized probability distribution) that simply gives you a `def apply(a: A): Double`. And now you want to be able to define joint _densities_ instead. It is fairly straight-forward to define it mathematically (density over `A and B` is simply the product of density over `A` and density over `B`), but `Density` is not an applicative. This is what prompted me to make this change.

Ideally I would write my tests in terms of `Density` but that's not a data type in cats, so I decided to go with `Semigroup` instead.

PR #248, opened 2020-08-30
Setup publishing with sbt-ci-release

joroKr21:
So that any maintainer can publish by pushing a tag.
Also fix semiauto docs and delete 2.11 specific code.

@kailuowang you would have to follow the steps here if we want to do this: https://github.com/olafurpg/sbt-ci-release

PR #258, opened 2020-10-24
Setup scalafmt

joroKr21:


jatcwang:
Hi @joroKr21, I'm planning to switch this over to github actions. Do you think we can get this in before I start to avoid the merge conflicts? (and I see you've already done some of the necessary upgrades such as sbt 1.4). Thanks :)

joroKr21:
Something's wrong with the `validate` command though - it's not working

joroKr21:
@jatcwang done ✅ Thanks for picking this up! There is also #248 but we can ignore it because I would rather start from scratch on top of GitHub actions.

jatcwang:
Thanks @joroKr21

PR #279, opened 2020-12-02
Switch Show and PrettyShow from Typeable to ClassTag

### LINKED ISSUE #267: Errors deriving show with generic case classes
In Scala 2.13.3, with kittens 2.2.0 ([scastie link](https://scastie.scala-lang.org/2YDxLDqrTlyvd84uFT50lw)):
```scala
import scala.annotation.nowarn
import cats.{Show, derived}
import com.comcast.ip4s.{Cidr, IpAddress}

case class PortForwarding[+IP <: IpAddress](
  protocol: String,
  internalIpAddress: IP,
  internalPort: Int,
  externalPort: Int,
)

case class Route[+IP <: IpAddress](destination: Cidr[IP], nexthop: IP)

implicit def show1[IP <: IpAddress: Show]: Show[PortForwarding[IpAddress]] = derived.semiauto.show
implicit def show2[IP <: IpAddress: Show]: Show[PortForwarding[IP]] = derived.semiauto.show
implicit def show3[IP <: IpAddress: Show]: Show[Route[IpAddress]] = derived.semiauto.show
implicit def show4[IP <: IpAddress: Show]: Show[Route[IP]] = derived.semiauto.show

```
`show1` compiles successfully. The `show2`, `show3`, and `show4` all output an error like:
```
Could not derive an instance of Show[A] where A = Playground.PortForwarding[IP].
Make sure that A has a Typeable instance and satisfies one of the following conditions:
  * it is a case class where all fields have a Show instance
  * it is a sealed trait where all subclasses have a Show instance
```

I assume this is a bug, especially since `show1` compiles but `show3` does not.

Removing the variance annotations produces the same result.

The shows for `IpAddress` and `Cidr` are defined:
```scala
package com.comcast.ip4s
object IpAddress {
  implicit def show[A <: IpAddress]: Show[A] = Show.fromToString[A]
}
object Cidr {
  implicit def show[A <: IpAddress]: Show[Cidr[A]] = Show.fromToString[Cidr[A]]
}
```

joroKr21:
We're not using any special features of `Typeable`
(e.g. we are not showing the type parameters)
that are not provided by `ClassTag`.

But it is confusing to users when there is no `Typeable` instance.
Fixes #267

### DIFF
--- a/core/src/main/scala/cats/derived/show.scala
+++ b/core/src/main/scala/cats/derived/show.scala
@@ -3,9 +3,10 @@ package derived
 
 import shapeless._
 import shapeless.labelled._
-import util.VersionSpecific.{OrElse, Lazy}
+import util.VersionSpecific.{Lazy, OrElse}
 
 import scala.annotation.implicitNotFound
+import scala.reflect.ClassTag
 
 /** Due to a limitation in the way Shapeless' `describe` is currently
   * implemented, `Show` can't be derived for ADTs which are _both_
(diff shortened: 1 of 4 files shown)


kailuowang (approved):
Lgtm

PR #330, opened 2021-05-11
Derive Functor, Foldable, Traverse

TimWSpence:


TimWSpence:
@joroKr21 opening to get some feedback :)

core/src/main/scala-3.0.0-RC3/cats/derived/foldable.scala
```diff
@@ -0,0 +1,59 @@
+package cats.derived
+
+import cats.{Eval, Foldable}
+import shapeless3.deriving.{K1, Continue}
+
+object foldable extends FoldableDerivation
+
+trait ProductFoldable[T[x[_]] <: Foldable[x], F[_]] extends Foldable[F]:
+
+  val inst: K1.ProductInstances[T, F]
```
joroKr21, line 10: Hmm I realise now we have a choice - classes, traits, traits with parameters (I think the last one will be most flexible)

core/src/main/scala-3.0.0-RC3/cats/derived/foldable.scala
```diff
@@ -0,0 +1,59 @@
...
+    new ProductFoldable[Foldable, F]{
+        val inst: K1.ProductInstances[Foldable, F] = summon[K1.ProductInstances[Foldable, F]]
+      }
+
+  given coproductFoldable[F[_]](using inst: => K1.CoproductInstances[Foldable, F]): Foldable[F] =
+    new CoproductFoldable[Foldable, F]{
+        val inst: K1.CoproductInstances[Foldable, F] = summon[K1.CoproductInstances[Foldable, F]]
+      }
+
+  given Foldable[Id] with
```
joroKr21, line 51: Isn't this provided by cats?

core/src/main/scala-3.0.0-RC3/cats/derived/foldable.scala
```diff
@@ -0,0 +1,59 @@
...
+
+object foldable extends FoldableDerivation
+
+trait ProductFoldable[T[x[_]] <: Foldable[x], F[_]] extends Foldable[F]:
+
+  val inst: K1.ProductInstances[T, F]
+
+  def foldLeft[A, B](fa: F[A], b: B)(f: (B, A) => B): B =
+    inst.foldLeft[A, B](fa)(b)(
+      [t[_]] => (acc: B, fd: T[t], t0: t[A]) => Continue(fd.foldLeft(t0, acc)(f))
```
joroKr21, line 14: Does `Continue` give us stack safety?

core/src/main/scala-3.0.0-RC3/cats/derived/functor.scala
```diff
@@ -0,0 +1,33 @@
+package cats.derived
+
+import cats.Functor
+import shapeless3.deriving.K1
+
+object functor extends FunctorDerivation
+
+trait ProductFunctor[T[x[_]] <: Functor[x], F[_]] extends Functor[F]:
```
joroKr21, line 8:
```suggestion
trait GenericFunctor[T[x[_]] <: Functor[x], F[_]] extends Functor[F]:
```

core/src/main/scala-3.0.0-RC3/cats/derived/traverse.scala
```diff
@@ -0,0 +1,60 @@
...
+    new ProductTraverse[Traverse, F]{
+        val inst: K1.ProductInstances[Traverse, F] = summon[K1.ProductInstances[Traverse, F]]
+      }
+
+  given coproductTraverse[F[_]](using inst: => K1.CoproductInstances[Traverse, F]): Traverse[F] =
+    new CoproductTraverse[Traverse, F]{
+        val inst: K1.CoproductInstances[Traverse, F] = summon[K1.CoproductInstances[Traverse, F]]
+      }
+
+  given Traverse[Id] with
```
joroKr21, line 43: Again I think this is provided by cats

core/src/main/scala-3.0.0-RC3/cats/derived/functor.scala
```diff
@@ -0,0 +1,33 @@
...
+      new ProductFunctor[Functor, F]{
+        val inst: K1.Instances[Functor, F] = summon[K1.Instances[Functor, F]]
+      }
+
+  given functorGen[F[_]](using inst: => K1.ProductInstances[Functor, F]): Functor[F] =
+    new ProductFunctor[Functor, F]{
+        val inst: K1.Instances[Functor, F] = summon[K1.Instances[Functor, F]]
+      }
+
+  given Functor[Id] with
```
joroKr21, line 29: Do we need this?

joroKr21 (commented):
This is amazing! WDYT about traits with `using` parameters? That seems the most flexible to me. I realise the classes I started with will not scale.

Are the `Id` instances needed? Does it not compile without them?

And the last question I'm wondering is about stack safety but we can leave it for later or decide we don't want to bother.

joroKr21:
If you rebase CI should pass now - I kept the Scala 2 sources on the same branch so that we have a reference point and tests to port - we can delete them later.

core/src/main/scala-3.0.0-RC3/cats/derived/foldable.scala
milessabin: It does. Imperative loop under the hood.

TimWSpence:
> This is amazing! WDYT about traits with `using` parameters? That seems the most flexible to me. I realise the classes I started with will not scale.
> 
> Are the `Id` instances needed? Does it not compile without them?
> 
> And the last question I'm wondering is about stack safety but we can leave it for later or decide we don't want to bother.

Good question about `using` params. I tried that before but at that point I thought that they also needed to be by name, which isn't allowed. But it seems that they don't so I'll try that again 👍 

Also good question about the `Id` instances. I forgot that cats also defines `Id` as a type alias rather than some kind of newtype so it might work - I'll try it 👍 

The actual shapeless 3 implementations are all loops (eg https://github.com/milessabin/shapeless/blob/shapeless-3/modules/deriving/src/main/scala/shapeless3/deriving/erased.scala#L148) so I believe stack-safety isn't an issue, unless I've missed something!

core/src/main/scala-3.0.0-RC3/cats/derived/foldable.scala
joroKr21: Maybe I'm missing something but I don't see how this can be stack safe due to alternating calls to `ProductInstances.foldLeft` and `CoproductInstances.fold` 🤔

milessabin: Hmm ... yes you're right.

TimWSpence: Good point! 🤔

joroKr21:
I'm gonna merge this so that we can iterate faster on top

joroKr21 (approved):

PR #336, opened 2021-05-14
Derive Eq, PartialOrder, Order and Hash

TimWSpence:


TimWSpence:
@joroKr21 this looks vaguely correct at least! Thought it would be good to get the structure in place at least

core/src/main/scala-3.0.0-RC3/cats/derived/hash.scala
```diff
@@ -0,0 +1,36 @@
...
+
+object hash extends HashDerivation
+
+trait ProductHash[T[x] <: Hash[x], A](using inst: K0.ProductInstances[T, A])
+  extends ProductEq[T, A], Hash[A]:
+
+  def hash(x: A): Int = {
+    val (hash, len) = inst.foldLeft[(Int, Int)](x)((MurmurHash3.productSeed, 0))(
+      [t] => (acc: (Int, Int), h: T[t], t0: t) => Continue((MurmurHash3.mix(acc._1, h.hash(t0)), acc._2 + 1))
+    )
```
joroKr21, line 16: I would rather do two iterations without allocating tuples

core/src/main/scala-3.0.0-RC3/cats/derived/partialOrder.scala
```diff
@@ -0,0 +1,37 @@
...
+        val cmp = ord.partialCompare(t0, t1)
+        Complete(cmp != 0)(cmp)(acc)
+      }
+    )
+
+trait CoproductPartialOrder[T[x] <: PartialOrder[x], A](using inst: K0.CoproductInstances[T, A])
+    extends PartialOrder[A]:
+
+  def partialCompare(x: A, y: A): Double =
+    //TODO is this correct? I _think_ it must be to be consistent with Order
```
joroKr21, line 23: Hmm in Kittens 2 we have the `PartialOrder` where two different subclasses are considered not comparable. I never thought about type class coherence in this case. But OTH it's way more intuitive to me that we can't compare them 🤔

joroKr21 (commented):
I think we should avoid allocations in `ProductHash`. Regarding `PartialOrder` - not sure, but I lean towards keeping the 2.0 semantics.

core/src/main/scala-3.0.0-RC3/cats/derived/hash.scala
joroKr21: Oh, even better: restrict `A <: Product` and use `productArity` directly

TimWSpence: Oh good call! 👍

core/src/main/scala-3.0.0-RC3/cats/derived/partialOrder.scala
TimWSpence: Yeah, I guess coherence only really demands that `partialCompare(x,y) != Nan ==> compare(x,y).toDouble == partialCompare(x,y)`. Or at least I would hope that that is the case.

However I can't think of a particularly good reason why we shouldn't compare ordinals when that is considered valid for deriving `Order` instances? Happy to trust your judgement on it though!

joroKr21: To me it's the other way around - why should we assume that we can compare two different branches of an ADT? We can make it work for `Order` via `ordinal` but it's kinda arbitrary.

core/src/main/scala-3.0.0-RC3/cats/derived/hash.scala
joroKr21: Hmm I gather that `A <: Product` doesn't quite work so we need `A <:< Product`.
I wonder if we can avoid the type casts somehow.

core/src/main/scala-3.0.0-RC3/cats/derived/partialOrder.scala
TimWSpence: Yep, fair enough 👍 Gives another reason to derive `PartialOrder` as well - if you want that behaviour instead of the arbitrary comparison of ordinals

core/src/main/scala-3.0.0-RC3/cats/derived/hash.scala
TimWSpence: I wouldn't want to guarantee that we can't 😂 But I couldn't see how to get `derived` to work, where we need `A <: Product` only if `Mirror[A]` is a `Product` and not a `Sum`

joroKr21 (approved):
Awesome 👏

TimWSpence:
AFAICT Github Actions just exploded ("internal errors") rather than it being a genuine  build failure? But I might have missed something

joroKr21:
Yes it has problems right now: https://www.githubstatus.com/

PR #340, opened 2021-05-19
Derive Show with field labels

TimWSpence:


core/src/main/scala-3.0.0-RC3/cats/derived/show.scala
```diff
@@ -0,0 +1,31 @@
...
+
+object show extends ShowDerivation
+
+trait ShowDerivation:
+
+  extension (F: Show.type)
+    inline def derived[A](using gen: K0.Generic[A]): Show[A] = gen.derive(productShow, coproductShow)
+
+  given productShow[A](using inst : => K0.ProductInstances[Show, A], l: Labelling[A]): Show[A] with
+    def show(a: A) = {
```
joroKr21, line 14: Let's use a SAM function instead (it doesn't generate anonymous classes)

core/src/main/scala-3.0.0-RC3/cats/derived/show.scala
```diff
@@ -0,0 +1,31 @@
...
+      var idx = 0
+      val x = inst.foldLeft[StringBuilder](a)(new StringBuilder(s"${l.label}("))(
+        [t] => (acc: StringBuilder, sh: Show[t], t0: t) => {
+          //elemLabels is backed by an array so this should be fast
+          val res = Continue(acc.append(s"${l.elemLabels(idx)}=${sh.show(t0)}, "))
+          idx = idx + 1
+          res
+          }
+      )
+      x.delete(x.length - 2, x.length).append(")")
```
joroKr21, line 24: I think this is not safe for empty products

core/src/main/scala-3.0.0-RC3/cats/derived/show.scala
```diff
@@ -0,0 +1,31 @@
...
+      x.toString()
+    }
+
+  given coproductShow[A](using inst: => K0.CoproductInstances[Show, A]): Show[A] with
+    def show(a: A) = inst.fold[String](a)(
```
joroKr21, line 29: Same here - let's use a SAM function

joroKr21 (changes requested):
Really nice 👍 See my comments for some minor issues.

core/src/main/scala-3.0.0-RC3/cats/derived/show.scala
TimWSpence: 🤦 good point!

core/src/main/scala-3.0.0-RC3/cats/derived/show.scala
TimWSpence: You mean define an explicit `ProductShow` trait like we've done for the other typeclasses?

joroKr21: I mean you can just use a function:
`given ... : Show[A] = a => ...`

core/src/main/scala-3.0.0-RC3/cats/derived/show.scala
```diff
@@ -0,0 +1,34 @@
...
+      val x = inst.foldLeft[StringBuilder](a)(new StringBuilder(s"${l.label}("))(
+        [t] => (acc: StringBuilder, sh: Show[t], t0: t) => {
+          //elemLabels is backed by an array so this should be fast
+          val res = Continue(acc.append(s"${l.elemLabels(idx)}=${sh.show(t0)}, "))
+          removeTrailingComma = true
+          idx = idx + 1
+          res
+          }
+      )
+      if(removeTrailingComma) {x.delete(x.length - 2, x.length)}
```
joroKr21, line 26: Isn't this just
```suggestion
      if (i > 0) {x.delete(x.length - 2, x.length)}
```

core/src/main/scala-3.0.0-RC3/cats/derived/show.scala
joroKr21: Isn't this just
```suggestion
      if (idx > 0) {x.delete(x.length - 2, x.length)}
```

TimWSpence: I should just give up on programming today 😂 Thanks

joroKr21 (approved):
Perfect 👍

PR #341, opened 2021-05-19
Derive Commutative{Semigroup, Monoid}

TimWSpence:


joroKr21 (approved):
Really nice and way easier than the Scala 2 version 😄

PR #345, opened 2021-05-21
Setup build for Scala 3

joroKr21:
Also use the constant type directly for brevity


TimWSpence (approved):
Nice! Hadn't noticed that Shapeless exports `Const` 🤦

TimWSpence (approved):

PR #346, opened 2021-05-21
Derive Contravariant and Invariant

TimWSpence:


joroKr21 (approved):
You need to resolve conflicts, but otherwise LGTM 👍

PR #347, opened 2021-05-21
Reducible derivation

joroKr21:
Also modifies `Functor`, `Foldable` and `Traverse` derivation for consistency:
 * Each type class has a `Derived*` equivalent to host the given scope
 * `const` instances live to the companion object
 * `delegated` instances via `export` replace `OrElse` from Kittens 2
 * `composed` instances also via `Delegated` in the companion object
 * Unfortunately we need `*LowPriority` scope (couldn't make it work by adding `DummyImplicit`)


core/src/test/scala-3/cats/derived/FoldableTests.scala
```diff
@@ -2,7 +2,7 @@ package cats.derived
 
 import cats.Foldable
 import cats.derived.all._
-import cats.derived.all.given
+import cats.instances.all._
```
TimWSpence, line 5: Does import given only work with instances defined with `given`, rather than the legacy `implicit` used in cats?

TimWSpence (approved):
Very nice!

core/src/test/scala-3/cats/derived/FoldableTests.scala
joroKr21: I don't know 🤔 - that import shouldn't be necessary anymore

TimWSpence: 👍 Did your cats change to put `Id` instances into given scope get released then?

joroKr21: No - but I think _some_ instances were already in implicit scope.
My change was to extend this to all instances.

TimWSpence: 🤦 😂

PR #349, opened 2021-05-25
Deep and Coproduct Empty derivation

joroKr21:
It looks like we will need `OrElse` again.


joroKr21:
> Why do we need the OrElse machinery?

So that we can have "deep" derivation that respects existing instances. You can see it in action in Kittens 2. But it doesn't quite work as expected.

TimWSpence:
> > Why do we need the OrElse machinery?
> 
> So that we can have "deep" derivation that respects existing instances. You can see it in action in Kittens 2. But it doesn't quite work as expected.

Oh of course! 🤦 I guess otherwise the given resolution would always pick our generic instance because it's directly in scope?

joroKr21:
> I guess otherwise the given resolution would always pick our generic instance because it's directly in scope?

Or we make it not `given` but then we can't support "deep" derivation, meaning if I say `Foo derives Bar` then if possible I should get generated instances for all subtypes and fields recursively.

PR #353, opened 2021-06-01
semiauto and auto imports

TimWSpence:


core/src/main/scala-3/cats/derived/instances.scala
```diff
@@ -27,3 +28,6 @@ private[derived] trait Instances1:
   given [X](using X: Semigroup[X]): SemigroupK[Const[X]] with
     def combineK[A](x: Const[X][A], y: Const[X][A]): Const[X][A] =
       X.combine(x, y)
+
+  given [X](using X: Empty[X]): EmptyK[Const[X]] with
```
joroKr21, line 32: 👍

core/src/main/scala-3/cats/derived/package.scala
```diff
@@ -0,0 +1,44 @@
...
+  MonoidDerivation,
+  MonoidKDerivation,
+  OrderDerivation,
+  PartialOrderDerivation,
+  ReducibleDerivation,
+  SemigroupDerivation,
+  SemigroupKDerivation,
+  ShowDerivation,
+  TraverseDerivation,
+  Instances:
```
joroKr21, line 25: I think we should remove those traits and move the definitions here (or top level in the `derived` package)

core/src/main/scala-3/cats/derived/package.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def foldable[F[_]](using ev: DerivedFoldable[F]): Foldable[F] = ev.instance
+
+  def functor[F[_]](using ev: DerivedFunctor[F]): Functor[F] = ev.instance
+
+  def reducible[F[_]](using ev: DerivedReducible[F]): Reducible[F] = ev.instance
+
+  def traverse[F[_]](using ev: DerivedTraverse[F]): Traverse[F] = ev.instance
```
joroKr21, line 33: Those should be inline similar to the extension methods

core/src/main/scala-3/cats/derived/package.scala
```diff
@@ -0,0 +1,44 @@
...
+
+object auto:
+
+  given [F[_]](using ev: DerivedFoldable[F]): Foldable[F] = ev.instance
+
+  given [F[_]](using ev: DerivedFunctor[F]): Functor[F] = ev.instance
+
+  given [F[_]](using ev: DerivedReducible[F]): Reducible[F] = ev.instance
+
+  given [F[_]](using ev: DerivedTraverse[F]): Traverse[F] = ev.instance
```
joroKr21, line 44: Those should be inline as well

joroKr21 (changes requested):
WDYT about my suggestions?

core/src/main/scala-3/cats/derived/package.scala
TimWSpence: Yeah maybe top level so you import `cats.derived.given` for using `deriving` and `cats.derived.auto.given` for magic

core/src/main/scala-3/cats/derived/package.scala
TimWSpence: Sorry, I probably need to look at the opaque type change in more detail. Why is `inline` necessary?

core/src/main/scala-3/cats/derived/package.scala
joroKr21: It would be `import cats.derived.*` though because the extension methods are not given.
Or we could put them in a `given` specifically for that purpose 🤔

core/src/main/scala-3/cats/derived/package.scala
joroKr21: Because `type DerivedFoldable[F[_]] = Derived[Foldable[F]]` and `object DerivedFoldable` are not actually companions and you would have to `import DerivedFoldable.given` - but the inline method does a trick and imports this scope before calling `summonInline[DerivedFoldable[F]]`.

core/src/main/scala-3/cats/derived/package.scala
```diff
@@ -0,0 +1,44 @@
...
+  OrderDerivation,
+  PartialOrderDerivation,
+  ReducibleDerivation,
+  SemigroupDerivation,
+  SemigroupKDerivation,
+  ShowDerivation,
+  TraverseDerivation,
+  Instances:
+
+  inline def foldable[F[_]](using ev: DerivedFoldable[F]): Foldable[F] = ev.instance
```
joroKr21, line 27: Sorry, this is what I meant:
```suggestion
  inline def foldable[F[_]]: Foldable[F] = DerivedFoldable[F]
```

core/src/main/scala-3/cats/derived/package.scala
```diff
@@ -0,0 +1,44 @@
...
+  inline def functor[F[_]](using ev: DerivedFunctor[F]): Functor[F] = ev.instance
+
+  inline def reducible[F[_]](using ev: DerivedReducible[F]): Reducible[F] = ev.instance
+
+  inline def traverse[F[_]](using ev: DerivedTraverse[F]): Traverse[F] = ev.instance
+
+
+object auto:
+
+  inline given [F[_]](using ev: DerivedFoldable[F]): Foldable[F] = ev.instance
```
joroKr21, line 38:
```suggestion
  inline given [F[_]]: Foldable[F] = DerivedFoldable[F]
```

core/src/main/scala-3/cats/derived/package.scala
TimWSpence: Ah! 👍 thanks

core/src/test/scala-3/cats/derived/FunctorTests.scala
```diff
@@ -1,7 +1,7 @@
 package cats.derived
 
 import cats.Functor
-import cats.derived.all._
+import cats.derived.given
```
joroKr21, line 4: Wait, are you telling me that this imports also extension methods 🤔

core/src/main/scala-3/cats/derived/package.scala
TimWSpence: I think an extension automatically goes into given scope? Or at least I just changed the tests to use `cats.derived.given` and that worked

core/src/test/scala-3/cats/derived/FunctorTests.scala
TimWSpence: I believe so!

TimWSpence: Argh no!!! It's because the tests are in the same package 😂

TimWSpence: When we're porting the tests, maybe we should put them in a different package to make sure we pick stuff like this up?

core/src/main/scala-3/cats/derived/package.scala
joroKr21: Neat 👍

core/src/test/scala-3/cats/derived/FunctorTests.scala
joroKr21: I guess so

joroKr21 (approved):

PR #355, opened 2021-06-02
Use summonFrom to prioritize DerivedEmpty instances

bplommer:


joroKr21:
Does it change anything semantically or is it just to avoid the low priority class?

bplommer:
Just to avoid the low priority class. I think it expresses the intention a lot more clearly.

joroKr21:
Thanks, that's a really good way to prioritise givens

PR #356, opened 2021-06-02
Further simplify Empty derivation

bplommer:
This reduces indirection but I think also makes it clearer what's going on - I found it confusing that `derive` takes `Generic` as an argument but then delegates to `product` which takes `ProductInstances`.


joroKr21 (approved):
LGTM

PR #357, opened 2021-06-02
Fix infinite loop in derived Empty

bplommer:
The `@main` would previously hang. This needs some cleaning up but seems to fix it.


joroKr21:
Superseded by #359

PR #359, opened 2021-06-02
Generic Derived opaque type for given priority

joroKr21:
cc @bplommer - it doesn't work with invariant `OrElse` either 🤔


bplommer:
Opaque types don't seem to behave how you'd expect with respect to variance - this compiles:

```scala
opaque type Foo[A] = A
object Foo:
  def apply[A](a: A): Foo[A] = a

val a: Foo[1] = Foo(1)
val b: Foo[Int] = a
```
Changing `Foo[A]` to `Foo[-A]` also makes no difference

joroKr21:
I think this finally works

TimWSpence (approved):
This is pretty neat!

So the `summonFrom` in `OrInstances`  is how we assign precedent to existing instances over derived ones?

joroKr21:
> So the summonFrom in OrInstances is how we assign precedent to existing instances over derived ones?

exactly 👍

PR #362, opened 2021-06-04
Port functor tests

TimWSpence:


TimWSpence:
Dependeny on https://github.com/typelevel/kittens/pull/353

TimWSpence:
Currently blocked on https://github.com/lampepfl/dotty/issues/13332

joroKr21:
@TimWSpence sorry for force push but we shouldn't modify the Scala 2 files 🙈 

I don't think we should wait for the resolution of the Scala 3 bug - we can make progress in the mean time. I made `TestDefns` an object, removed the stack safety test and made `AndChar` use `Int` instead of `Long` - it's the only one still throwing a `ClassCastException` and I have no clue why ...

PR #401, opened 2021-11-02
Fix semigroup hierarchy derivation for recursive types

TimWSpence:
Fix semigroup hierarchy derivation for recursive types

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ val catsVersion = "2.6.1"
 val disciplineMunitVersion = "1.0.9"
 val kindProjectorVersion = "0.13.0"
 val shapeless2Version = "2.3.7"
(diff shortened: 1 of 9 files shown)


joroKr21:
@TimWSpence did the CI hang? I try to cancel it but it doesn't work. I think this is blocked on typelevel/shapeless-3#40

TimWSpence:
🤦 thanks @joroKr21, I think I did see that issue and then completely forgot about it. I'll have a proper look through it later

joroKr21 (approved):
I think we should give the same treatment to all product instances

joroKr21:
Oh no I think it's still hanging 🤔

joroKr21:
Superseded by #455

PR #427, opened 2022-01-13
Bump Scala and Cats and comment out tests that don't work

joroKr21:


TimWSpence (approved):
Thanks for this! So many FIXMEs 😭 

We're just going to 3.1.0 now then and accepting the regressions for the moment?

joroKr21:
> We're just going to 3.1.0 now then and accepting the regressions for the moment?

I don't see another choice - Scala native only works with 3.1 and there are other bug fixes we do want.
I think so far the problem is only with `Reducible` and `NonEmptyTraverse` which are not used that often.

gvolpe:
@joroKr21 think we can get another milestone published including these fixes? 🙏🏽

joroKr21:
Yes, I think just pushing a tag should do it - let me do that today

PR #429, opened 2022-01-21
Migrate to sbt-typelevel-ci-release

armanbilge:


armanbilge:
(This one's also ready for review btw :)

.sbtrc
```diff
@@ -1 +1 @@
-alias boot = ;reload ;project kittens ;iflast shell
+alias boot = ;reload ;project / ;iflast shell
```
joroKr21, line 1: No idea what this is for

joroKr21 (approved):
Thanks and sorry for the delay - I'm really busy these days 😅

.sbtrc
armanbilge: @joroKr21 This `.sbtrc` file? Or the change I made here?

This was a very strange setup and if you don't know what it is I think we should just remove it.

joroKr21: Yes the `.sbtrc` file

PR #433, opened 2022-01-31
Port Show with tests

andrzejressel:
The only thing that does not work is implicit `Show[Address]` when `implicit` is changed to `given`.


joroKr21 (approved):
Nice, thanks 👍 This brings us one step closer to completing Scala 3 support 🥳

PR #439, opened 2022-02-07
Port Applicative to Dotty

andrzejressel:
Requires Scala [3.1.2-RC1](https://github.com/lampepfl/dotty/releases/tag/3.1.2-RC1)

### DIFF
--- a/core/src/main/scala-3/cats/derived/DerivedApplicative.scala
+++ b/core/src/main/scala-3/cats/derived/DerivedApplicative.scala
@@ -1,0 +1,44 @@
+package cats.derived
+
+import shapeless3.deriving.{Const, K1}
+import cats.{Applicative, Monoid}
+
+import scala.compiletime.*
+import shapeless3.deriving.{Continue, K0, Labelling}
+
+import scala.annotation.implicitNotFound
+import scala.deriving.Mirror
+
+@implicitNotFound("""Could not derive an instance of Applicative[F] where F = ${F}.
+Make sure that F[_] satisfies one of the following conditions:
+  * it is a constant type λ[x => T] where T: Monoid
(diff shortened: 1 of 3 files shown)


andrzejressel:
I was some tuple fixes in 3.1.1 - maybe that will help

andrzejressel:
I've commented out Tuple examples and added comment, so it can be merged as is, without any waiting.

core/src/main/scala-3/cats/derived/DerivedApplicative.scala
```diff
@@ -0,0 +1,41 @@
...
+    def ap[A, B](ff: T)(fa: T) = T.combine(ff, fa)
+
+  given [F[_], G[_]](using F: Applicative[F], G: Applicative[G]): DerivedApplicative[[x] =>> F[G[x]]] =
+    F.compose(G)
+
+  given [F[_]](using inst: => K1.ProductInstances[Or, F]): DerivedApplicative[F] =
+    given K1.ProductInstances[Applicative, F] = inst.unify
+    new Product[Applicative, F] {}
+
+  trait Product[T[x[_]] <: Applicative[x], F[_]](using inst: K1.ProductInstances[T, F]) extends Applicative[F]:
```
joroKr21, line 38: Actually the idea is that you can reuse your `Apply` implementation so you should add `extends DerivedApply.Product[T, F]` here

andrzejressel: Yeah, I've realized that after I finished Apply :D

joroKr21:
It's not a big deal about the tests - it's just satisfying if it works because it makes it easier to add another test. But if it doesn't work, oh well we lived with it in Scala 2.

joroKr21 (approved):
LGTM 👍

PR #440, opened 2022-02-07
Port Order to Dotty

andrzejressel:


andrzejressel:
After merge I'll propably make every derivation lazy

joroKr21:
> After merge I'll propably make every derivation lazy

WDYM? 🤔

andrzejressel:
I basically mean replacing:
- `using inst: K0.ProductInstances[Or, A]` with `using inst: => K0.ProductInstances[Or, A]`
- `using using inst: K0.CoproductInstances[Or, A]` with `using inst: => K0.CoproductInstances[Or, A]`

And the same for `K1` and `Instances[Or, A]`.

joroKr21:
> So recursive data classes will work properly.

I think @TimWSpence tried that in #401 but for some reason it's still not working properly

andrzejressel:
Yeah, i saw that. I'll check if scala/shapeless update will fix that.

core/src/main/scala-3/cats/derived/DerivedOrder.scala
```diff
@@ -0,0 +1,45 @@
...
+type DerivedOrder[A] = Derived[Order[A]]
+object DerivedOrder:
+  type Or[A] = Derived.Or[Order[A]]
+
+  inline def apply[A]: Order[A] =
+    import DerivedOrder.given
+    summonInline[DerivedOrder[A]].instance
+
+  @targetName("product")
+  given [A](using inst: => K0.ProductInstances[Or, A]): DerivedOrder[A] =
```
joroKr21, line 20: Let's just use the name
```suggestion
  given product[A](using inst: => K0.ProductInstances[Or, A]): DerivedOrder[A] =
```

core/src/main/scala-3/cats/derived/DerivedOrder.scala
```diff
@@ -0,0 +1,45 @@
...
+    given K0.ProductInstances[Order, A] = inst.unify
+    new Product[Order, A] {}
+
+  @targetName("coproduct")
+  given [A](using inst: => K0.CoproductInstances[Or, A]): DerivedOrder[A] =
```
joroKr21, line 25:
```suggestion
  given coproduct[A](using inst: => K0.CoproductInstances[Or, A]): DerivedOrder[A] =
```

core/src/test/scala-3/cats/derived/OrderSuite.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import cats.Order
+import cats.kernel.laws.discipline.{OrderTests, SerializableTests}
+import org.scalacheck.Arbitrary
+
+class OrderSuite extends KittensSuite {
+  import OrderSuite._
+  import TestDefns._
+
+  def testOrder(context: String)(using
```
joroKr21, line 27: Same here - we can simplify by making this `inline` like with your `Apply` PR

joroKr21 (commented):
Thanks again - this is amazing! ❤️

core/src/test/scala-3/cats/derived/OrderSuite.scala
andrzejressel: Here we don't need any Isomorphism and inline trick works great 👍

core/src/main/scala-3/cats/derived/DerivedOrder.scala
andrzejressel: Fixed

joroKr21 (approved):
Thank you! One step closed to feature parity 👍

PR #448, opened 2022-02-15
Port EmptyK to dotty

andrzejressel:


core/src/main/scala-3/cats/derived/DerivedEmptyK.scala
```diff
@@ -0,0 +1,43 @@
...
+    import DerivedEmptyK.given
+    summonInline[DerivedEmptyK[F]].instance
+
+  given [T](using T: Empty[T]): DerivedEmptyK[Const[T]] = new EmptyK[Const[T]]:
+    def empty[A] = T.empty
+
+  given or[F[_], G[_]](using F: Or[F]): DerivedEmptyK[[x] =>> F[G[x]]] = new EmptyK[[x] =>> F[G[x]]]:
+    def empty[A] = F.unify.empty
+
+  given pureor[F[_], G[_]](using NotGiven[Or[F]])(using F: Pure[F], G: Or[G]): DerivedEmptyK[[x] =>> F[G[x]]] =
```
joroKr21, line 32: Interesting why is `NotGiven[Or[F]]` a separate parameter list? I also wonder - did you have to specify the given name because it's ambiguous with `or`? 🤔

core/src/main/scala-3/cats/derived/DerivedEmptyK.scala
```diff
@@ -0,0 +1,43 @@
...
+    new EmptyK[[x] =>> F[G[x]]]:
+      def empty[A] = F.pure(G.unify.empty)
+
+  given [F[_]](using inst: K1.ProductInstances[Or, F]): DerivedEmptyK[F] =
+    given K1.ProductInstances[EmptyK, F] = inst.unify
+    new Product[EmptyK, F] {}
+
+  trait Product[T[x[_]] <: EmptyK[x], F[_]](using
```
joroKr21, line 40: Do you think you can replicate `DerivedEmpty.coproduct` to get n `DerivedEmptyK.coproduct`? That was not possible with shapeless 2 but I think it should be possible with shapeless 3.

joroKr21 (commented):
Hey @andrzejressel sorry for being MIA so long. Thanks for helping, much appreciated 👍 
I have a few comments.

core/src/main/scala-3/cats/derived/DerivedEmptyK.scala
andrzejressel: Mostly because when one given instance is assigned to variable all of then have to be. And it kinda looked weird to me to assign Notgiven to variable.

andrzejressel: To be honest I don't know if I prefer this or some tricks with inheritance to set priority to implicits.

joroKr21: Ahh I see - the priority should already prefer the one with the less amount of `using` dependencies

joroKr21 (approved):

PR #475, opened 2022-05-24
Port contravariant to new scheme

TimWSpence:
Several of the scala 2 test cases are currently commented out with TODOs pointing to open issues


core/src/main/scala-3/cats/derived/DerivedContravariant.scala
```diff
@@ -0,0 +1,35 @@
...
+object DerivedContravariant:
+  type Or[F[_]] = Derived.Or[Contravariant[F]]
+  inline def apply[F[_]]: Contravariant[F] =
+    import DerivedContravariant.given
+    summonInline[DerivedContravariant[F]].instance
+
+  given [T]: DerivedContravariant[Const[T]] = new Contravariant[Const[T]]:
+    def contramap[A, B](fa: T)(f: B => A): T = fa
+
+  given [F[_], G[_]](using F: DerivedFunctor.Or[F], G: Or[G]): DerivedContravariant[[x] =>> F[G[x]]] =
```
joroKr21, line 25: I wonder if we can now support the other composition as well which is `F: Contravariant` and `G: Functor` - I think it was ambiguous in Scala 2.

TimWSpence: Yeah I was just doing `Invariant` and realized there are missing cases. I assume so 👍

TimWSpence: Literally came here to check if that instance existed and found your comment 😂

joroKr21: What do you mean that `Invariant` needs the same treatment?
About making a separate issue - feel free 👍

TimWSpence: Yeah, there are two more instances we could define for `Invariant`. Opened https://github.com/typelevel/kittens/issues/479 to track it

Issue #477, opened 2022-05-25
[Scala 3] Consistency on requiring `Or` instances for givens


TimWSpence:
eg in `DerivedFunctor` we have
```scala
  given [F[_], G[_]](using F: Or[F], G: Or[G]): DerivedFunctor[[x] =>> F[G[x]]] =
    F.unify.compose(G.unify)

  given [F[_], G[_]](using F: Contravariant[F], G: Contravariant[G]): DerivedFunctor[[x] =>> F[G[x]]] =
    F.compose(G)
```

and in `DerivedSemigroupK` we have
```scala
  given [F[_], G[_]](using F: Or[F]): DerivedSemigroupK[[x] =>> F[G[x]]] =
    F.unify.compose[G]

  given [F[_], G[_]](using N: NotGiven[Or[F]], F: DerivedApply.Or[F], G: Or[G]): DerivedSemigroupK[[x] =>> F[G[x]]] =
    new SemigroupK[[x] =>> F[G[x]]]:
      final override def combineK[A](x: F[G[A]], y: F[G[A]]): F[G[A]] = F.unify.map2(x, y)(G.unify.combineK(_, _))
```

These should probably be consistent. I would have thought that we would need the `Or` version but things seem to work fine without it

PR #484, opened 2022-05-30
Replace all usages of implicit with given for Scala 3

joroKr21:
Also use braceless syntax in more places.

Partially addresses #477


joroKr21:
Superseded by #485

PR #486, opened 2022-05-30
Require Or instances for givens

TimWSpence:
Resolves https://github.com/typelevel/kittens/issues/477.

However, this feature doesn't really seem to work as hoped (possibly related to https://github.com/typelevel/kittens/issues/473).

```scala
  case class In[A](value: A => Unit) derives Contravariant
  case class Out[A](value: A => Unit) derives Contravariant
  case class N[A](value: In[Out[A]]) derives Functor
```

works fine. But

```scala
  case class In[A](value: A => Unit)
  case class Out[A](value: A => Unit)
  case class N[A](value: In[Out[A]]) derives Functor
```

does not


joroKr21 (approved):
Yes I think it's related. Let's merge this now but we have to investigate before releasing 3.0

PR #490, opened 2022-05-31
Derive ShowPretty

TimWSpence:
This one is a bit different because we derive `ShowPretty` if we don't have a `Show` instance, rather than a `ShowPretty` instance.

Resolves https://github.com/typelevel/kittens/issues/488

### DIFF
--- a/core/src/main/scala-3/cats/derived/DerivedShowPretty.scala
+++ b/core/src/main/scala-3/cats/derived/DerivedShowPretty.scala
@@ -1,0 +1,65 @@
+package cats.derived
+
+import cats.Show
+import shapeless3.deriving.{Continue, K0, Labelling}
+
+import scala.annotation.implicitNotFound
+import scala.compiletime.*
+import scala.deriving.Mirror
+
+trait ShowPretty[A] extends Show[A]:
+  def showLines(a: A): List[String]
+  def show(a: A): String = showLines(a).mkString(System.lineSeparator)
+
+object ShowPretty:
+  inline def apply[A](using A: ShowPretty[A]): A.type = A
+
+@implicitNotFound("""Could not derive an instance of ShowPretty[A] where A = ${A}.
+Make sure that A satisfies one of the following conditions:
+  * it is a case class where all fields have a Show instance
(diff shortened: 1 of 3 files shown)


core/src/main/scala-3/cats/derived/DerivedShowPretty.scala
```diff
@@ -0,0 +1,71 @@
...
+@implicitNotFound("""Could not derive an instance of ShowPretty[A] where A = ${A}.
+Make sure that A satisfies one of the following conditions:
+  * it is a case class where all fields have a Show instance
+  * it is a sealed trait where all subclasses have a Show instance""")
+type DerivedShowPretty[A] = Derived[ShowPretty[A]]
+object DerivedShowPretty:
+  opaque type Or[A] = A => List[String]
+  object Or extends OrInstances:
+    def apply[A](instance: A => List[String]): Or[A] = instance
+    extension [A](or: Or[A]) def apply(a: A): List[String] = or(a)
```
joroKr21, line 26: Why is this so different than the rest of derivations? 🤔 
I thought we would either derive `ShowPretty` directly or use the regular scheme.

TimWSpence: `ShowPretty` is weird because the expectation seems to be that you don't define instances yourself. It's driven off [`Show OrElse DerivedShowPrettty`](https://github.com/typelevel/kittens/blob/master/core/src/main/scala/cats/derived/showPretty.scala#L36). This is why we have `NotGiven[Show]` above as well - otherwise the `auto` tests fail because they don't honour existing `Show` instances

TimWSpence: Also I guess we would be creating ambiguous implicits with `auto` otherwise - `ShowPretty <: Show` so we shouldn't create an instance of `ShowPretty` if there is already a `Show` instance for a given type

joroKr21: Oh yeah I forgot about `Show OrElse DerivedShowPretty` 👍

PR #498, opened 2022-06-10
Test scala 3 enums

TimWSpence:
Test derivation of typeclasses for scala 3 `enum`s. Addresses https://github.com/typelevel/kittens/issues/495


joroKr21 (approved):
Cool, more tests are always better

PR #501, opened 2022-06-16
Cross-build for Native

armanbilge:
Also various build updates while I'm here.


armanbilge:
Woops, JVM tests are failing, probably because of the Cats v2.8.0 update? 😬 Hopefully someone else can look into it 🙏 

```
[error] Failed: Total 4078, Failed 5, Errors 0, Passed 4073
[error] Failed tests:
[error]         cats.derived.EmptySuite
[error]         cats.derived.PureSuite
[error] (coreJVM / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 9 s, completed Jun 16, 2022, 10:31:17 PM
```

joroKr21:
There was no issue with #502

joroKr21:
Ah sorry, this is for Scala 3 🤔

joroKr21:
@armanbilge I think it's because typelevel/cats#4060 changed the implicit not found message

armanbilge:
Ahh, phew! So not really a regression after all.

joroKr21 (approved):
Thanks 👍

joroKr21:
Is it just me or adding Scala native to some projects makes it impossible to run `test` locally? 🤔

armanbilge:
Hm, how is that? You may need to install LLVM and stuff.

Alternatively, rootJVM/test works fine 👍

joroKr21:
> Hm, how is that? You may need to install LLVM and stuff.

I have it installed but the test/compile basically hangs indefinitely because it's running out of memory 🤔

armanbilge:
Ooh, yes it can suck up a lot of memory 😬 maybe we should add a .jvmopts file

joroKr21:
> Ooh, yes it can suck up a lot of memory 😬 maybe we should add a .jvmopts file

But ScalaJS doesn't have the memory issue although it also has to call an external compiler.

joroKr21:
Also CI doesn't have the same issue - I suspect it has to do with the number of available cores (i.e. when you have many cores available it tries to do too much in parallel and doesn't consider the available memory).

armanbilge:
I'm not sure. This is only a single module project, I usually see those parallelism issues in multi module projects. Even in CI actually

armanbilge:
Ah, question: what happens if you just run rootNative/test in a new sbt session? The problem could be doing test which does JS and Native in parallel

joroKr21:
> Ah, question: what happens if you just run rootNative/test in a new sbt session? The problem could be doing test which does JS and Native in parallel

Yeah that works in Kittens, but not in Cats - I guess because it's multi-module 😄

armanbilge:
Yes, this is why in CI the linking step is run separately and sbt is restarted before running the tests. Only way I found to make it reliable ...

PR #513, opened 2022-09-07
Upgrade to Scala 3.2

TimWSpence:
Contains a load of fixes that we need, particularly for Mirror synthesis

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.15, 2.13.8, 3.1.2]
+        scala: [2.12.15, 2.13.8, 3.2.0]
         java: [temurin@8]
         project: [rootJS, rootJVM, rootNative]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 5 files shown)


joroKr21:
You need to run `githubWorkflowGenerate`

TimWSpence:
> You need to run `githubWorkflowGenerate`

Sorry, forgot to check the build before I logged off yesterday 🤦

joroKr21:
> Does scala native not support 3.2 yet?

I think the version that supports is 0.4.7
We should really merge this branch to `master` and release it soon.
Is there something still missing? I guess more tests with enums 🤔

joroKr21 (approved):

TimWSpence:
@joroKr21 thanks for fixing the rest of the dependencies! Sorry I hadn't got round to it yet

joroKr21:
> @joroKr21 thanks for fixing the rest of the dependencies! Sorry I hadn't got round to it yet

No worries, I was thinking if we should force the update to 3.2 but I think with all the bugs around mirrors we really should.

TimWSpence:
> > @joroKr21 thanks for fixing the rest of the dependencies! Sorry I hadn't got round to it yet
> 
> No worries, I was thinking if we should force the update to 3.2 but I think with all the bugs around mirrors we really should.

Yeah I think so too. https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html seems to recommend that everyone upgrade as well, although I don't know if people are following that advice in general. At some point should we move to sbt-typelevel to let it handle a lot of the build pain for us?

joroKr21:
We are using `sbt-typelevel`

PR #516, opened 2022-09-13
Add tests for all derived instances

### LINKED ISSUE #425: [Scala 3] - Wildcard derivation import not working
Super simple reproduction:

```scala
import cats.Show
// import cats.derived.semiauto.{ derived, product } // this works
// import cats.derived.semiauto.{ given, * } // this works (as per the 3rd comment)
import cats.derived.semiauto.* // does not work

enum TradeAction derives Show:
  case Ask, Bid
```

It fails with the following error:

```
no implicit argument of type cats.Show[(Playground.TradeAction.Ask : Playground.TradeAction)] was found

The following import might fix the problem:

  import cats.derived.semiauto.product
```

https://scastie.scala-lang.org/0q4XaUbJTqibfe78MuyzAA

---

Not sure whether this has already been discussed or not but I figured it might be better to keep track of it right here.

joroKr21:
Unfortunately there is some code duplication because I didn't find a way to make this generic.
But the boilerplate is minimal because we just wrap existing ADTs into a sort of newtype case class.

Closes #425 by making sure it works for all type classes


core/src/test/scala-3/cats/derived/ADTs.scala
```diff
@@ -0,0 +1,378 @@
...
+        Arbitrary.arbitrary[Int].map(LeafI.apply)
+      )
+    )
+
+    given Cogen[EnumK0] = Cogen[Either[String, Int]].contramap {
+      case LeafS(s) => Left(s)
+      case LeafI(i) => Right(i)
+    }
+
+  enum EnumK1[A]:
```
TimWSpence, line 43: Was this meant to be Covariant?

core/src/test/scala-3/cats/derived/KittensSuite.scala
```diff
@@ -16,18 +16,23 @@
...
+import org.scalacheck.{Arbitrary, Cogen}
 import org.scalacheck.Test.Parameters
 
+import scala.deriving.Mirror
 import scala.quoted.*
 
 /** An opinionated stack of traits to improve consistency and reduce boilerplate in Kittens tests. Note that unlike the
   * corresponding CatsSuite in the Cat project, this trait does not mix in any instances.
   */
-abstract class KittensSuite extends KittensSuite.WithoutEq, TestEqInstances
```
TimWSpence, line 30: What are these instances used for? I thought we were defining explicit instances for everything in `ADTS`?

core/src/test/scala-3/cats/derived/ADTs.scala
```diff
@@ -0,0 +1,378 @@
...
+
+package cats.derived
+
+import cats.{Eq, Eval}
+import cats.syntax.all.given
+import org.scalacheck.rng.Seed
+import org.scalacheck.{Arbitrary, Cogen, Gen}
+import scala.annotation.tailrec
+
+object ADTs:
```
TimWSpence, line 25: So are we dropping testing for sealed traits in favour of enums? I had assumed that we would test both (although that sucks 😂 )

TimWSpence (approved):
This is awesome! Thanks so much!!

core/src/test/scala-3/cats/derived/KittensSuite.scala
joroKr21: We have the `EqInstances` separately to be able to test `Eq` without interference 🐔 🥚

core/src/test/scala-3/cats/derived/ADTs.scala
joroKr21: We're not dropping anything 🤔 - we just add more tests

core/src/test/scala-3/cats/derived/ADTs.scala
joroKr21: I think it doesn't matter

gvolpe:
Thank you for all this work 😍 ! New milestone coming out soon?

core/src/test/scala-3/cats/derived/ADTs.scala
TimWSpence: Oh so we are! 🤦

TimWSpence:
> Thank you for all this work 😍 ! New milestone coming out soon?

We're aiming for at least an RC if not full release ASAP :)

PR #517, opened 2022-09-13
[Scala 2] Add Record and Union instances

### LINKED ISSUE #60: [Scala 2] Instances for Records and Unions
Kittens allows us to work with raw HLists and Coproducts:
```scala
import cats.instances.all._
import cats.derived.MkMonoid._
import cats.syntax.monoid._

1 :: "a" :: HNil |+| 2 :: "b" ::HNil // ok
```
However, we can't work with raw Records and Unions:
```scala
'k ->> 1 :: HNil |+| 'k ->> 2 :: HNil // no Monoid instance
```
The same goes for other type classes as well. It mostly boils down to defining an instance for `FieldType[K, V]`. Are there any objections to adding them?

joroKr21:
Boils down to supporting `FieldType` instances. Closes #60

### DIFF
--- a/core/src/main/scala-2/cats/derived/applicative.scala
+++ b/core/src/main/scala-2/cats/derived/applicative.scala
@@ -17,8 +17,9 @@
 package cats
 package derived
 
+import cats.derived.util.VersionSpecific.OrElse
 import shapeless._
-import util.VersionSpecific.OrElse
+import shapeless.labelled._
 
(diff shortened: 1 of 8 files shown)


joroKr21:
Looks like it doesn't work for Scala 2.12

joroKr21:
This can move forward once we drop Scala 2.12

PR #582, opened 2023-07-05
Future-proof auto implementation

julienrf:
For context see lampepfl/dotty#18130.

We put fewer constraints to program elaboration by making some methods `transparent`.

I have checked locally that the PR compiles with Scala 3.3.2-RC1-bin-20230703-0a21ecf-NIGHTLY.


joroKr21:
It needs a format

joroKr21:
So I'm not sure about the problem in Dotty, but the changes here - just need to run scalafmt

julienrf:
I am not sure adding `transparent` here is the right way to go.

PR #626, opened 2023-12-17
Strict semiauto derivation for Scala 3

joroKr21:


mrdziuban:
This looks great, thank you @joroKr21! Have you given any thought to implementing the same in scala 2?

joroKr21:
> This looks great, thank you @joroKr21! Have you given any thought to implementing the same in scala 2?

Scala 2 is much more tricky. We can't have inline imports, so I'm not sure.

mrdziuban:
Thank you for getting this done @joroKr21! Do you have any sense of when you might plan to release a new version with it included?

joroKr21:
> Do you have any sense of when you might plan to release a new version with it included?

Yeah, I just want to write up documentation for the README and we can release it, so real soon.

PR #795, opened 2026-06-16
Fix/issue 476 stack safety

### LINKED ISSUE #476: [Scala 3] derived instances are not stack-safe
The following derivations are not stack-safe:
- `Functor`
- `Contravariant`

This is because the shapeless 3 `ErasedXXXInstances#map` has mutual recursion between the `Product` and `Coproduct` instances when dealing with recursive sum types like `List`

Arolg:
Closes #476.

Fixes the StackOverflowError reported in #476 for derived Functor/Contravariant (and other typeclasses) on recursive ADTs. Default derivation (semiauto.* / Eq.derived / strict) keeps the existing fast direct-recursion path, so existing users see no behavior or performance change. A new opt-in entry point cats.derived.stackSafe.* provides Eval-trampolined variants for the 12 K0/K1 typeclasses; one import on the user side is enough to switch. Bifunctor/Bifoldable/Bitraverse keep fast direct recursion only (the two-given-set design doesn't fit the current K2-coproduct resolver in shapeless3 — happy to discuss a follow-up). All 12 081 existing tests pass; StackSafetySuite now covers all 12 K0/K1 instances via the opt-in entry point.

### DIFF
--- a/core/src/main/scala-3/cats/derived/DerivedBifoldable.scala
+++ b/core/src/main/scala-3/cats/derived/DerivedBifoldable.scala
@@ -74,30 +74,9 @@ object DerivedBifoldable:
   given [F[_, _]](using => CoproductInstances[Bifoldable |: Derived, F]): DerivedBifoldable[F] =
     Strict.coproduct
 
-  private[derived] trait Safe[F[_, _]] extends Bifoldable[F]:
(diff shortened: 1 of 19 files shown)


joroKr21:
The code needs to be formatted

